PR #6, opened 2019-05-16
Refactor api

wussler:


emersion:
It would be nice to have a summary of the changes (and even nicer to have smaller commits with clean changesets).

wussler:
> It would be nice to have a summary of the changes (and even nicer to have smaller commits with clean changesets).

It's still in progress, it will soon be added to the summary and the changelog.
Smaller commits are not possible because most of the logic is changed, introducing new message classes.

Next step is to create some helpers for the most common encryption/decryption/signing methods.
Examples have not been reviewed yet.

emersion:
>It's still in progress, it will soon be added to the summary and the changelog.

Nice!

>Smaller commits are not possible because most of the logic is changed, introducing new message classes.

+917 −1,140 commits are hard to review. Making clearly separated commits helps a lot, and is often possible by forcing yourself to do one thing at a time (e.g. introduce one type per commit, or introduce the new type in one commit and start using it in subsequent commits). I agree it does require to put some thought into it before starting to the work, however for a security-related library IMHO it's worth it.

>Next step is to create some helpers for the most common encryption/decryption/signing methods.
Examples have not been reviewed yet.

Looking forward to having a look at the API.

L11R:
It also would be nice to refactor `GenerateKey()` method. Currently you even cannot set different name and email, because username used to generate email, lol. Also there is no expiration date and so on.

wussler:
@L11R I changed the name/email parameters to be more flexible, Expiration date is for another PR, as I would consider better adding another function with explicit expiration for it.

PR #17, opened 2019-06-29
Fix compilation for gomobile iOS

wussler:
Create struct ClearTextMessage
Create struct SignatureVerificationError
Drop struct Verification

Alter functions:
* (*Keyring) VerifyDetached
* (*Keyring) Decrypt

Create functions: 
* NewClearTextMessage
* NewClearTextMessageFromArmored
* (*ClearTextMessage) GetArmored
* (*ClearTextMessage) GetString
* (*ClearTextMessage) GetBinary
* (*ClearTextMessage) GetSignature

Drop functions:
* ReadClearSignedMessage
* ArmorClearSignedMessage

### DIFF
--- a/ProposalChanges.md
+++ b/ProposalChanges.md
@@ -63,19 +63,29 @@ type pgpKeyObject struct {
 }
 ```
 
-## Dropped
-### Signature
-```
-type Signature struct {
-	md *openpgp.MessageDetails
-}
-```
 ### SignedString
 ```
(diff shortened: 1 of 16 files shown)


L11R:
So we did it just because iOS cannot handle returned structs? Don't you think it was the hasty decision? I personally prefer dedicated methods for gomobile with some suffix/preffix, or even dedicated helper package in the library.

PR #22, opened 2019-08-22
Edits for bridge

wussler:
Add some functions and tests for bridge development


crypto/keyring.go
```diff
@@ -296,6 +296,11 @@ func (keyRing *KeyRing) ReadFromJSON(jsonData []byte) (err error) {
 	return keyRing.newKeyRingFromPGPKeyObject(keyObjs)
 }
 
+// UnmarshalJSON is an alias for ReadFromJSON to implement the encoding/json.Unmarshaler interface
+func (keyRing *KeyRing) UnmarshalJSON(jsonData []byte) (err error) {
+	return keyRing.ReadFromJSON(jsonData)
+}
```
twiss: This is a bit nitpicky but - I don't think we should really have two public APIs for this. It sort of seems to imply that this is an important / good way to serialize / deserialize a list of keys, which I'm not so sure about. It seems better to have a list of armored keys or multiple keys armored together. So I know golang's names are weird but I'd just revert to having `UnmarshalJSON` for compatibility and nothing more.

crypto/keyring_message.go
```diff
@@ -89,10 +91,12 @@ func asymmetricEncrypt(data []byte, publicKey *KeyRing, privateKey *KeyRing, isB
 		FileName: "",
 	}
 
+	// We want to use the first entity only for encryption, the API will already return them in the correct order
+	// The first key is meant to be the primary
```
twiss: This seems suspect; sometimes we need to encrypt to multiple recipients, no? If the API returns multiple keys for a single recipient, it's the responsibility of the app to pass only the first one, not the responsibility of this library.

crypto/message.go
```diff
@@ -221,7 +223,19 @@ func (msg *PGPSplitMessage) GetBinaryKeyPacket() []byte {
 	return msg.KeyPacket
 }
 
+// GetBinaryPackets returns the unarmored binary joined packets as a []byte
+func (msg *PGPSplitMessage) GetBinaryPackets() []byte {
+	return append(msg.KeyPacket , msg.DataPacket...)
```
twiss:
```suggestion
	return append(msg.KeyPacket, msg.DataPacket...)
```

crypto/message.go
```diff
@@ -221,7 +223,19 @@ func (msg *PGPSplitMessage) GetBinaryKeyPacket() []byte {
 	return msg.KeyPacket
 }
 
+// GetBinaryPackets returns the unarmored binary joined packets as a []byte
+func (msg *PGPSplitMessage) GetBinaryPackets() []byte {
```
twiss: `GetBinary`? For consistency with `PGPMessage`

PR #23, opened 2019-08-23
Hotfix to prevent use-after-free bug on iOS

ZortacDev:
The NewPGPMessage() function is currently broken when GopenPGP is used with Gomobile on iOS as the memory pointed to in the PGPMessage struct is erroneously freed by the Swift garbage collector. This error is discussed in golang/go#33745 (additional discussion in mssun/passforios#289).

This hotfix copies the encrypted data into memory controlled by Go's garbage collector in the NewPGPMessage() function to prevent issues caused by the Swift garbage collector's erroneous free and restores the functionality of NewPGPMessage().


ZortacDev:
This PR is intended only as a temporary fix and should be reverted once the upstream issue in Gomobile is fixed.

PR #27, opened 2019-09-20
update build script and other minor changes

zhj4478:
*update the build script to be more generic.
*update some code correct by lint
*separate the functions to a different file and added the build tag to avoid go-mobile to compile unsupported functions
*renamed ios helper to mobile helper because it also used in android
*added a few more helper functions for mobile


wussler:
Could you please change the name of the helper `helper_none_mobile.go`?
Helpers are not only specific to mobile :)

zhj4478:
> Could you please change the name of the helper `helper_none_mobile.go`?
> Helpers are not only specific to mobile :)
I know it is not only for mobile. there is a mobile.go file. `helper_none_mobile.go` this file contains the function doesn't support on gomobile
I don't have a better name, suggestions?

PR #31, opened 2019-12-04
Openpgp security update (V2)

wussler:
*TL;DR: To have a keyring, you un-armor each key separately, unlock it, add it to the keyring. Can't add locked keys to keyring. Remember to erase keyrings when done. Can now use session key with a high-level API directly.*

0. Security means no more strings. All strings for secretes will be replaced by byte arrays, and strings will not be used internally.
1. Since we can't have a single endpoint where all the keys get unlocked at the same time, so a new object was introduced: the Key.  Mixed locked/unlocked keyrings, are no longer possible, and to add keys to a keyring they must be already unlocked.
How does it work?
```go
privateKey, _ := NewKeyFromArmored("---- BEGIN PGP PRIVATE KEY BLOCK ----- ....")
unlockedKey, _ := privateKey.Unlock(passphrase)
ok, _ = unlockedKey.Check()
if !ok {
  panic("Private key does not match public key")
}
keyring, _ = NewKeyRing(unlockedKey)
// Unarmor, check and unlock further keys
keyring.AddKey(key2)
// Do stuff
keyring.Decrypt(...)
// Clear keyring, will also clear all unlocked keys
keyring.ClearPrivateParams()
```
 2. new function :sparkles: (bool) Key#Check. It will check the private key against the public key.
 3. more new functions :sparkles: (bool) keyring#ClearPrivateParams(), Key#ClearPrivateParams(). It will overwrite the keyring memory, and check if it was correctly erased. if not will return false.
If you run it on the keyring will be executed for all contained keys.
(See point (1)).
 4. SymmetricKey is now SessionKey. The encrypt and decrypt function will operate directly on key packets. To lock stuff with password (i.e. encrypt with a symmetric key packet) there are helpers functions. 
The interface is now:
```go
var sessionKey, _ = GenerateSessionKey()
var message = NewPlainMessageFromString("The secret code is... 1, 2, 3, 4, 5")
// Encrypt data with session key
dataPacket, _ := sessionKey.Encrypt(message)
// Decrypt data with session key
decrypted, _ := sessionKey.Decrypt(dataPacket)
```
Regarding password protected messages:
```go
var message = NewPlainMessageFromString("The secret code is... 1, 2, 3, 4, 5")
var pass = []byte("hunter2")
// Encrypt data with password
encrypted, _ := EncryptMessageWithPassword(message, pass)
// Decrypt data with password
decrypted, _ := DecryptMessageWithPassword(encrypted, pass)
```
 5. Update of the underlying crypto lib


v2/crypto/key_clear.go
```diff
@@ -0,0 +1,128 @@
...
+
+	return num > 0
+}
+
+func (key *Key) clearPrivateWithSubkeys() (num int) {
+	num = 0
+	if key.entity.PrivateKey != nil {
+		err := clearPrivateKey(key.entity.PrivateKey.PrivateKey)
+		if err == nil {
+			num += 1
```
houlahaj: `golint` will complain here. Use `num++` instead

v2/crypto/sessionkey.go
```diff
@@ -0,0 +1,211 @@
...
+	}
+
+	sk = &SessionKey{
+		Key: r,
+		Algo: algo,
+	}
+	return sk, nil
+}
+
+// RandomToken generated a random key with the default cipher
```
houlahaj: I feel this comment is wrong?

v2/crypto/key_clear.go
houlahaj: (same for below)

v2/crypto/key.go
```diff
@@ -0,0 +1,393 @@
...
+
+	err = lockedKey.entity.PrivateKey.Encrypt(passphrase)
+	if err != nil {
+		return nil, errors.New("gopenpgp: error in locking key: " + err.Error())
+	}
+
+	for _, sub := range lockedKey.entity.Subkeys {
+		if sub.PrivateKey != nil {
+			if err := sub.PrivateKey.Encrypt(passphrase); err != nil {
+				return nil, errors.New("gopenpgp: error in locking key: " + err.Error())
```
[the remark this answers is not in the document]
wussler: Will change to errors.Wrap :+1:

v2/crypto/sessionkey.go
wussler: Yup, changed

v2/crypto/key.go
```diff
@@ -0,0 +1,393 @@
...
+
+// IsPrivate returns true if the key is private
+func (key *Key) IsPrivate() bool {
+	return key.entity.PrivateKey != nil
+}
+
+// IsLocked checks if a private key is locked
+func (key *Key) IsLocked() (bool, error) {
+	if key.entity.PrivateKey == nil {
+		return true, errors.New("gopenpgp: a public key can not be locked")
```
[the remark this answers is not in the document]
wussler: Amended

crypto/key.go
```diff
@@ -4,194 +4,390 @@ import (
...
+// GenerateKey generates a key of the given keyType ("rsa" or "x25519").
+// If keyType is "rsa", bits is the RSA bitsize of the key.
+// If keyType is "x25519" bits is unused.
+func GenerateKey(name, email string, keyType string, bits int) (*Key, error) {
+	return generateKey(name, email, keyType, bits, nil, nil, nil, nil)
+}
+
+// --- Operate on key
+
+// Lock fully locks a copy of the keyring.
```
horejsek: Wrong comment.

crypto/key.go
```diff
@@ -4,194 +4,390 @@ import (
...
+		return nil, errors.New("gopenpgp: unable to lock key")
 	}
 
-	newEntity, err := openpgp.NewEntity(name, comments, email, cfg)
+	return lockedKey, nil
+}
+
+// Unlock fully unlocks a copy of the key
+func (key *Key) Unlock(passphrase []byte) (*Key, error) {
+	unlocked, err := key.IsLocked()
```
horejsek: This variable should be `isLocked`.

crypto/key.go
```diff
@@ -4,194 +4,390 @@ import (
...
-				return "", err
+	err = unlockedKey.entity.PrivateKey.Decrypt(passphrase)
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: error in unlocking key")
+	}
+
+	for _, sub := range unlockedKey.entity.Subkeys {
+		if sub.PrivateKey != nil {
+			if err := sub.PrivateKey.Decrypt(passphrase); err != nil {
+				return nil, errors.Wrap(err, "gopenpgp: error in unlocking key")
```
horejsek: `error in unlocking subkey`

README.md
```diff
@@ -245,16 +247,17 @@ if err == nil {
...
 -----END PGP PRIVATE KEY BLOCK-----` // encrypted private key
 const passphrase = "LongSecret"
 
-var message = NewPlainMessage(data)
+var message = crypto.NewPlainMessage(data)
 
-signingKeyRing, err := crypto.BuildKeyRingArmored(privkey)
-signingKeyRing.UnlockWithPassphrase(passphrase) // if private key is locked with passphrase
+privateKeyObj, err := crypto.NewKeyFromArmored(privkey)
+unlockedKeyObj = privateKeyObj.Unlock(passphrase)
```
horejsek: :=

README.md
```diff
@@ -84,50 +81,42 @@ import "github.com/ProtonMail/gopenpgp/crypto"
...
+decrypted, err := DecryptMessageWithPassword(encrypted, password)
 
 //Original message in decrypted.GetBinary()
 ```
 
 ### Encrypt / Decrypt with PGP keys
 
-```go
-import "github.com/ProtonMail/gopenpgp/helper"
+```gotemplate
```
horejsek: Why not simply go to have syntax highlight?

crypto/message.go
```diff
@@ -233,6 +233,10 @@ func (msg *PGPSplitMessage) GetArmored() (string, error) {
 	return armor.ArmorWithType(msg.GetBinary(), constants.PGPMessageHeader)
 }
 
+func (msg *PGPSplitMessage) GetPGPMessage() *PGPMessage {
```
horejsek: Missing doc string.

crypto/sessionkey.go
```diff
@@ -0,0 +1,211 @@
...
+type SessionKey struct {
+	// The decrypted binary session key.
+	Key []byte
+	// The symmetric encryption algorithm used with this key.
+	Algo string
+}
+
+var symKeyAlgos = map[string]packet.CipherFunction{
+	constants.ThreeDES:  packet.Cipher3DES,
+	constants.TripleDES: packet.Cipher3DES,
```
horejsek: What's the difference between ThreeDES and TripleDES?

crypto/sessionkey.go
```diff
@@ -0,0 +1,211 @@
...
+
+// GetCipherFunc returns the cipher function corresponding to the algorithm used
+// with this SessionKey.
+func (sk *SessionKey) GetCipherFunc() packet.CipherFunction {
+	cf, ok := symKeyAlgos[sk.Algo]
+	if ok {
+		return cf
+	}
+
+	panic("gopenpgp: unsupported cipher function: " + sk.Algo)
```
horejsek: Shouldn't be better to return error than panic?

crypto/sessionkey.go
```diff
@@ -0,0 +1,211 @@
...
+
+	encryptWriter, err := packet.SerializeSymmetricallyEncrypted(&encBuf, config.Cipher(), sk.Key, config)
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: unable to encrypt")
+	}
+
+	if algo := config.Compression(); algo != packet.CompressionNone {
+		encryptWriter, err = packet.SerializeCompressed(encryptWriter, algo, config.CompressionConfig)
+		if err != nil {
+			return nil, errors.Wrap(err, "gopenpgp: unable to encrypt")
```
horejsek: There is already `gopenpgp: unable to encrypt` few lines above. Would be good to have different message so it can be easier to spot from where the error is coming from. Also, I have feeling this is more `unable to serialize` than encrypt, anyway?

crypto/sessionkey.go
```diff
@@ -0,0 +1,211 @@
...
+			if err != nil {
+				return nil, errors.Wrap(err, "gopenpgp: unable to decrypt symmetric packet")
+			}
+
+		default:
+			return nil, errors.New("gopenpgp: invalid packet type")
+	}
+	_, err = decBuf.ReadFrom(decrypted)
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: unable to decrypt symmetric packet")
```
horejsek: `unable to read decrypted symmetric packet`?

crypto/key.go
```diff
@@ -4,194 +4,390 @@ import (
...
+			return false, nil
 		}
 	}
 
-	serialized := w.Bytes()
-	return armor.ArmorWithType(serialized, constants.PrivateKeyHeader)
+	return !key.entity.PrivateKey.Encrypted, nil
+}
+
+// Check verifies if the if the public keys match the private key parameters by signing and verifying
```
horejsek: if the if the :-)

go.mod
```diff
@@ -1,11 +1,12 @@
-module github.com/ProtonMail/gopenpgp
+module github.com/ProtonMail/gopenpgp/v2
```
cuthix: now I see.. OK.. so it seems that both options `.v2` and `/v2` are valid

I would expect that:
* `/v2` means that our repo has folder `/v2`
* `.v2` means you don't have the folder and you will not continue with `v1`

I don't think it's a rule so feel free to ignore all those comments

README.md
wussler: My IDE complained a lot about go not starting with `package`.
I used gotemplate that locally gave highlighting but no such errors.
Since it's not compatible with github I'll revert this change

crypto/sessionkey.go
wussler: The underlying crypto is the same, but some refer to it as threeDES ("3des"), other tripleDES ("tripledes")

crypto/sessionkey.go
wussler: Yeah, would be better, will amend the whole stack

crypto/sessionkey.go
wussler: `unable to serialize` is already a few lines under where we actually pass the data to serialize, i will go for `error in compression` :)

README.md
```diff
@@ -39,7 +39,7 @@ GO111MODULE=on go mod vendor
 
 ## Documentation
 
-https://godoc.org/github.com/ProtonMail/gopenpgp/crypto
+https://godoc.org/gopkg.in/ProtonMail/gopenpgp.v2/crypto
```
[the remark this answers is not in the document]
wussler: Nope, I wish.  I just used its API to provide versioned documentation.

wussler:
@cuthix 
> General comment: I would acitvate `golint` linter in CI. Mainly for documentation, because this is a public tool.
We now have a linter :sparkles:  (I changed the enabled modules a bit)

crypto/sessionkey.go
horejsek: I was interested why we have to keep both and Kubko explained it to me. Thanks!

crypto/keyring.go
```diff
@@ -347,12 +173,57 @@ func FilterExpiredKeys(contactKeys []*KeyRing) (filteredKeys []*KeyRing, err err
...
+
+	entities := make([]*openpgp.Entity, len(keyRing.entities))
+	for id, entity := range keyRing.entities {
+		var buffer bytes.Buffer
+		var err error
+
+		if entity.PrivateKey == nil {
+			err = entity.Serialize(&buffer)
+		} else {
+			err = entity.SerializePrivateNoSign(&buffer, nil)
```
twiss: Perhaps this code should be moved to the method on Key, and we should be calling that?
(I know, I know, it's not an API/interface comment.. feel free to ignore until later)

helper/ios_test.go
```diff
@@ -3,20 +3,22 @@ package helper
...
+	"github.com/ProtonMail/gopenpgp/v2/constants"
+	"github.com/ProtonMail/gopenpgp/v2/crypto"
 	"github.com/stretchr/testify/assert"
 )
 
 func TestIOSSignedMessageDecryption(t *testing.T) {
-	testPrivateKeyRing, _ := crypto.BuildKeyRingArmored(readTestFile("keyring_privateKey", false))
-	testPublicKeyRing, _ := crypto.BuildKeyRingArmored(readTestFile("mime_publicKey", false))
-
+	privateKey, _ := crypto.NewKeyFromArmored(readTestFile("keyring_privateKey", false))
```
zugzwang: We should avoid ignoring errors, even in test functions.

crypto/signature_collector.go
```diff
@@ -85,7 +85,8 @@ func (sc *SignatureCollector) Accept(
 			if err != nil {
 				return err
 			}
-			buffer, err = gomime.DecodeCharset(buffer, params)
+			mediaType, _, _ := mime.ParseMediaType(header.Get("Content-Type"))
```
zugzwang: It wouldn't hurt checking `ParseMediaType` error also. Since
```
 If there is an error parsing the optional parameter, the media type will be returned along with the error ErrInvalidMediaParameter.
```
and we're not passing any optional parameters, it is safe for now. But it would be safe to check the error in case we do pass parameters in the future or the error use in `mime` changes.

crypto/key.go
```diff
@@ -4,194 +4,390 @@ import (
...
+			fmt.Println("SubKey:" + hex.EncodeToString(subKey.PublicKey.Fingerprint[:]))
+		}
+	}
+	fmt.Println("PrimaryKey:" + hex.EncodeToString(key.entity.PrimaryKey.Fingerprint[:]))
+}
+
+// GetID returns the key ID
+func (key *Key) GetID() string {
+	return "0x" + strconv.FormatUint(key.entity.PrimaryKey.KeyId, 16)
+}
```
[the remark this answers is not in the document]
wussler: Ok, will rename it to `GetKeyID` and without `0x`

crypto/keyring.go
```diff
@@ -32,244 +24,85 @@ type Identity struct {
...
+// --- Extract info from key
 
-// BuildKeyRingNoError does not return error on fail
-func BuildKeyRingNoError(binKeys []byte) (keyRing *KeyRing) {
-	keyRing, _ = BuildKeyRing(binKeys)
-	return
+// CountEntities returns the number of entities in the keyring
+func (keyRing *KeyRing) CountEntities() int {
+	return len(keyRing.entities)
 }
```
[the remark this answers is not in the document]
wussler: Debug on gomobile, it has already been used for android in some cases, when checking if the correct number of decryption keys has been loaded.
In many scenarios, not being able to return arrays makes these compromises necessary

helper/ios_test.go
wussler: OK, added to plan for 2.0.1 under test improvements

crypto/signature_collector.go
wussler: Same as above, not being an API change will do in 2.0.1

crypto/keyring.go
wussler: Would you just clone it by cloning all the keys in the keyring, instead of having this "duplicate" code?

wussler: In any case it is postponed to 2.0.1 :)

crypto/key.go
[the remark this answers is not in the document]
wussler: Created `GetHexKeyID` and `GetKeyID`

PR #36, opened 2020-02-04
Fix TestMultipleKeyMessageEncryption

zugzwang:
When ciphertext data ends with a Symmetrically Encrypted Packet, the next subsequent call to `packet.Next()` after this packet tries to read an OpenPGP packet from the reader embedded in the Symmetrically Encrypted Packet, causing undesirable behaviour.

This tests now only checks (with 4 calls to `packet.Next()`) that there are 3 Encrypted Key Packets and one Symmetrically Encrypted Data Packet, plus correctness of decryption.

### DIFF
--- a/crypto/message_test.go
+++ b/crypto/message_test.go
@@ -3,7 +3,6 @@ package crypto
 import (
 	"bytes"
 	"encoding/base64"
-	"io"
 	"testing"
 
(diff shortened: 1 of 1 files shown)


wussler:
See https://github.com/ProtonMail/gopenpgp/pull/37

PR #39, opened 2020-03-19
doc: update README to use valid go.mod example syntax

whereswaldon:
This commit simply uses parenthesis in the example `go.mod` `require` block since curly braces are illegal there. If a potential user copies the text as the basis for starting a new `go.mod`, they'll get cryptic compilation errors.


twiss:
Thanks! :+1:

PR #41, opened 2020-04-06
Add SHA256 fingerprint utils and helpers

wussler:
Since the proven compromise of SHA1 hashes, this PR introduces a way to obtain SHA256 fingerprints for v4 keys too

### DIFF
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -3,6 +3,7 @@ package crypto
 import (
 	"bytes"
 	"crypto"
+	"crypto/sha256"
 	"encoding/hex"
 	"fmt"
 	"io"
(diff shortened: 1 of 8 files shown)


crypto/key.go
```diff
@@ -321,8 +322,24 @@ func (key *Key) GetFingerprint() string {
...
+	for _, sub := range key.entity.Subkeys {
+		fingerprints = append(fingerprints, hex.EncodeToString(computeSHA256Fingerprints(sub.PublicKey)))
+	}
+	return
+}
+
 // --- Internal methods
 
+// computeSHA256Fingerprints compute a fingerprint from a public key object
+func computeSHA256Fingerprints(pk *packet.PublicKey) []byte {
```
twiss, line 337: This shouldn't be plural fingerprints - maybe
```suggestion
func getSHA256FingerprintBytes(pk *packet.PublicKey) []byte {
```
?

crypto/key.go
```diff
@@ -14,9 +15,9 @@ import (
 	"github.com/ProtonMail/gopenpgp/v2/constants"
 	"github.com/pkg/errors"
 
-	"golang.org/x/crypto/openpgp"
+	openpgp "golang.org/x/crypto/openpgp"
 	xarmor "golang.org/x/crypto/openpgp/armor"
-	"golang.org/x/crypto/openpgp/packet"
+	packet "golang.org/x/crypto/openpgp/packet"
```
twiss, line 20: What does this do?

PR #44, opened 2020-04-24
Update lint

zugzwang:
Enable all linters of https://github.com/golangci/golangci-lint, maintain some exceptions. 

Enable linters `stylecheck` (replacement for golint) and `gocognit`  (complexity of functions), and refactor `SignatureCollector.Accept`.

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -33,7 +33,7 @@ func (e SignatureVerificationError) Error() string {
 // newSignatureFailed creates a new SignatureVerificationError, type SignatureFailed
 func newSignatureFailed() SignatureVerificationError {
 	return SignatureVerificationError{
-		constants.SignatureFailed,
+		constants.SIGNATURE_FAILED,
 		"Invalid signature",
 	}
 }
@@ -41,7 +41,7 @@ func newSignatureFailed() SignatureVerificationError {
 // newSignatureNotSigned creates a new SignatureVerificationError, type SignatureNotSigned
 func newSignatureNotSigned() SignatureVerificationError {
 	return SignatureVerificationError{
-		constants.SignatureNotSigned,
+		constants.SIGNATURE_NOT_SIGNED,
 		"Missing signature",
 	}
 }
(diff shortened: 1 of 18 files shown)


zugzwang:
My bad, I guess I linted locally with an older version of `golangci-lint`.
Re. constants, we could exclude those 4 constants but maintain the linter so that we won't have more like these in the future. Would you agree?

wussler:
Thanks! Looks good now :)

> Re. constants, we could exclude those 4 constants but maintain the linter so that we won't have more like these in the future. Would you agree?
I am not very happy to have mixed constant styles in a project, and we're a bit locked in this now.
I'd personally exclude the constant linter.

Any strong opinion @twiss?

zugzwang:
1/3 ground: I left the linter on, since it has other good checks, and excluded all `CamelCase` errors (`exclude ST1003`). 

Side comment, I don't think there's good reason to use those constants from the outside, maybe someone can prove me wrong.

zugzwang:
Elaborating a bit on my last comment, I believe these constants are used only as statuses. The user, when using `Decrypt` on a ciphertext missing a signature, only gets 
```
SignatureVerificationError{
    Status: 1  // internally called SIGNATURE_NOT_SIGNED
    Message: "Missing signature"
}
```

twiss:
> OTOH, a user could be checking something as follows: (...) If this seems reallistic, we can just leave those constants as they are (i.e. current version) and consider this out of the scope of this PR.

Hmhm, I would leave them as is for this reason indeed.

> I am not very happy to have mixed constant styles in a project, and we're a bit locked in this now.
I'd personally exclude the constant linter. Any strong opinion @twiss?

Yeah, I agree. If we want to change the constant style, we could do so in v3, perhaps.

crypto/signature.go
```diff
@@ -30,74 +31,73 @@ func (e SignatureVerificationError) Error() string {
...
-				md.SignatureError = nil
-			}
-		} else {
-			// verifyTime = 0: time check disabled, everything is okay
-			md.SignatureError = nil
-		}
+	if md.SignatureError != pgpErrors.ErrSignatureExpired {
+		return
+	}
+	if verifyTime <= 0 {
```
wussler, line 67: I know that this is the correct refactor, but I think we should stick to checking `verifyTime == 0`

Issue #35, opened 2020-01-22
Error when signing with unprotected armored key


dalai4git:
In [cleartext.go#L18](https://github.com/ProtonMail/gopenpgp/blob/3f33c71496d73010c679f4eff842f8cad6b439d2/helper/cleartext.go#L18) when using SignCleartextMessageArmored with an unprotected armored key an error occurs. This is due to the attempt to unlock the key even when the password is empty.

A fix would be to have a check and only do `unlockedKey, err := signingKey.Unlock(passphrase)` when the passphrase is not empty.

Issue #34, opened 2020-01-22
Error when doing symmetric encryption


dalai4git:
I am symmetrically encrypting a binary file with gopenpgp and decrypt it with gpg2 on linux. The file can be decrypted by gpg2, but is corrupted. The reason is the way it was encrypted. Specifically in [crypto/password.go#L109](https://github.com/ProtonMail/gopenpgp/blob/54f45d0471cf4b588a9cdc9ff013465d31196a01/crypto/password.go#L109) the line

	encryptWriter, err := openpgp.SymmetricallyEncrypt(&outBuf, password, nil, config)

means that the message is treated as text. I think this would need to be changed to

	hints := openpgp.FileHints{IsBinary: true}
	encryptWriter, err := openpgp.SymmetricallyEncrypt(&outBuf, password, &hints, config)

Issue #33, opened 2020-01-22
Bad header in armored signature


dalai4git:
When creating an armored signature in [crypto/message.go#L369](https://github.com/ProtonMail/gopenpgp/blob/54f45d0471cf4b588a9cdc9ff013465d31196a01/crypto/message.go#L369), the following header is used:

	str := "-----BEGIN PGP SIGNED MESSAGE-----\r\nHash:SHA512\r\n\r\n"

When validating the resulting message with gpg2 on linux an error is shown because it expects a space after the semicolon, i.e. `Hash: SHA512`.

PR #45, opened 2020-04-26
Fix various minor issues

wussler:
Closes https://github.com/ProtonMail/gopenpgp/issues/33
Closes https://github.com/ProtonMail/gopenpgp/issues/34
Closes https://github.com/ProtonMail/gopenpgp/issues/35

- Password encrypted binary files now have the correct flags
- Fixed missing space in `Hash` header of cleartext messages
- Providing empty passphrase does no longer throw an error when unlocking an unencrypted private key
- Improved memory zeroing in helpers

### DIFF
--- a/helper/sign_attachment.go
+++ b/helper/sign_attachment.go
@@ -30,24 +30,22 @@ func EncryptSignAttachment(
 		return nil, nil, nil, err
 	}
 
+	if unlockedKeyObj, err = privateKeyObj.Unlock(passphrase); err != nil {
+		return nil, nil, nil, err
+	}
+	defer unlockedKeyObj.ClearPrivateParams()
+
+	if privateKeyRing, err = crypto.NewKeyRing(unlockedKeyObj); err != nil {
+		return nil, nil, nil, err
+	}
+
 	if packets, err = publicKeyRing.EncryptAttachment(binMessage, fileName); err != nil {
 		return nil, nil, nil, err
 	}
 
-	if unlockedKeyObj, err = privateKeyObj.Unlock(passphrase); err != nil {
-		return nil, nil, nil, err
-	}
-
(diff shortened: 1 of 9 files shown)


helper/sign_attachment.go
```diff
@@ -30,21 +30,24 @@ func EncryptSignAttachment(
...
+	if privateKeyRing, err = crypto.NewKeyRing(unlockedKeyObj); err != nil {
+		unlockedKeyObj.ClearPrivateParams()
 		return nil, nil, nil, err
 	}
 
 	if signatureObj, err = privateKeyRing.SignDetached(binMessage); err != nil {
+		unlockedKeyObj.ClearPrivateParams()
 		return nil, nil, nil, err
 	}
+	unlockedKeyObj.ClearPrivateParams()
```
twiss, line 50: This is pretty nitpicky but since it's kinda cool I thought I'd mention it anyway:

Instead of repeating the `ClearPrivateParams` three times, you can do:
```go
	if unlockedKeyObj, err = privateKeyObj.Unlock(passphrase); err != nil {
		return nil, nil, nil, err
	}
	defer unlockedKeyObj.ClearPrivateParams()
```
to automatically call it when the function returns.

Maybe this is the way we should recommend to use this function in the readme and stuff :)

zugzwang: Sliding in to raise a question: Shouldn't `KeyRing.ClearPrivateParams` return an `ok bool`, as `Key.ClearPrivateParams`? 
In which case, usage could be recommended as
```
defer func() {
    if !unlockedKeyObj.ClearPrivateParams() {
        panic("Couldn't clear private params!")
    }
}()
```
(or something more appropriate)

wussler: If it returns false it means there was nothing more to clear, and IMHO is nothing worth panicking

I will introduce the defer and merge :)

PR #50, opened 2020-06-03
Clarify how to decrypt in README

klische:
It's not clear how to decrypt a binary file in the README


wussler:
Hi @klische, thank you for your contribution.
First of all, I'll take care of disabling golang 1.11 in CI, so that you can rebase on top.

Then, I'd like to ask for a change to this PR: instead of adding a whole new example, that implies redundancy, just add a comment to the previous example, on line 182, specifying that a `pgpMessage` can be generated with the function `crypto.NewPGPMessage([]byte)`.

Cheers

klische:
@wussler Thanks for taking a look at this PR. Fixed according to your recommendation.

PR #52, opened 2020-06-23
Reject signatures using insecure hash algorithms

twiss:
Reject message signatures using MD5, RIPEMD or SHA1.


crypto/signature.go
```diff
@@ -81,20 +89,19 @@ func processSignatureExpiration(md *openpgp.MessageDetails, verifyTime int64) {
...
+		len(verifierKey.entities) == 0 ||
+		len(verifierKey.entities.KeysById(md.SignedByKeyId)) == 0 {
+		return newSignatureNoVerifier()
 	}
-
-	return newSignatureNoVerifier()
+	if md.SignatureError != nil ||
+		md.Signature == nil ||
+		md.Signature.Hash < allowedHashes[0] ||
+		md.Signature.Hash > allowedHashes[len(allowedHashes)-1] {
```
zugzwang, line 101: This depends on the order of the hash listing on `x/crypto`. Not that it's going to change anytime soon but I would simply check against the `allowedHashes` global variable introduced in this commit, so that adding/removing hashes will be as simple as modifying it.

crypto/signature.go
```diff
@@ -81,20 +89,19 @@ func processSignatureExpiration(md *openpgp.MessageDetails, verifyTime int64) {
...
-		}
+	if !md.IsSigned ||
+		md.SignedBy == nil ||
+		len(verifierKey.entities) == 0 ||
+		len(verifierKey.entities.KeysById(md.SignedByKeyId)) == 0 {
+		return newSignatureNoVerifier()
 	}
-
-	return newSignatureNoVerifier()
+	if md.SignatureError != nil ||
```
zugzwang, line 98: Maybe an `&&` for extra security? Or is there a case where a `SignatureError` is given on a valid signature?

crypto/signature.go
twiss: Yeah. I kind of agree with you but this would introduce an extra loop (as there is no built-in function to check for inclusion in an array) and make this code roughly three times bigger. I think we can do that if we ever want to change `allowedHashes` to be a non-continuous range (which I find somewhat unlikely as usually the disallowed ones will be the weakest).

crypto/signature.go
twiss: The body of this if statement returns an error, so it errors if there is a `SignatureError` no matter what else happens.

twiss:
Yeah, I agree. I was kind of thinking we could ask them to check on the web to find out more, since there it also tells them which hash algorithm was used etc. But I've at least created a separate error for this to give some hint :)

PR #55, opened 2020-07-01
Add getEntity and getEcryptionKeyIDs functions to key and message types respectively

chesnokovilya:


crypto/key.go
```diff
@@ -354,6 +354,11 @@ func (key *Key) GetSHA256Fingerprints() (fingerprints []string) {
 	return
 }
 
+// GetEntity get xcrypto Entity object.
```
wussler, line 357:
```suggestion
// GetEntity gets x/crypto Entity object.
```

crypto/message.go
```diff
@@ -223,6 +223,24 @@ func (msg *PGPMessage) GetArmoredWithCustomHeaders(comment, version string) (str
 	return armor.ArmorWithTypeAndCustomHeaders(msg.Data, constants.PGPMessageHeader, version, comment)
 }
 
+// getEncryptionKeyIds Returns the key IDs of the keys to which the session key is encrypted.
+func (msg *PGPMessage) getEncryptionKeyIDs() []uint64 {
+	packets := packet.NewReader(bytes.NewReader(msg.Data))
+	var err error
```
wussler, line 229: Is it safe to ignore errors here?

crypto/message.go
```diff
@@ -223,6 +223,24 @@ func (msg *PGPMessage) GetArmoredWithCustomHeaders(comment, version string) (str
...
+		var p packet.Packet
+		if p, err = packets.Next(); err == io.EOF {
+			break
+		}
+		enc, ok := p.(*packet.EncryptedKey)
+		if ok {
+			ids = append(ids, enc.KeyId)
+		}
+	}
+	return ids
```
wussler, line 241: Shouldn't we issue an error in case IDs is empty?

crypto/message.go
chesnokovilya: yes, it is same pattern as in SeparateKeyAndData.

crypto/message.go
chesnokovilya: This is matter of preference, for example KeysById in x/crypto could return empty array of keys. In javascript this function could also return empty array. I could add an error, through probably bool flag is better.

CHANGELOG.md
```diff
@@ -16,6 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 (msg *PGPMessage) GetArmoredWithCustomHeaders(comment, version string) (string, error)
 ```
 
+- Now one could get key IDs of keys to which the session key is encrypted
+```go
+func (msg *PGPMessage) getEncryptionKeyIDs() ([]uint64, bool)
+```
+
+get x/crypto Entity (components of an OpenPGP key) from Key struct
```
wussler, line 24:
```suggestion
- Getter for the x/crypto Entity (internal components of an OpenPGP key) from Key struct
```

CHANGELOG.md
```diff
@@ -16,6 +16,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 (msg *PGPMessage) GetArmoredWithCustomHeaders(comment, version string) (string, error)
 ```
 
+- Now one could get key IDs of keys to which the session key is encrypted
```
wussler, line 19:
```suggestion
- Extraction of encryption key IDs from a PGP message, i.e. the IDs of the keys used in the encryption of the session key
```

PR #59, opened 2020-07-08
Encrypt session key to multiple keys in keyring.

chesnokovilya:
`EncryptSessionKey` now creates encrypted key packet for each valid key
in keyring.


twiss:
I don't really like the behavior of creating a session key packet for each _valid_ encryption key. IMO, if we're gonna create a session key packet for each key, we should error for invalid keys. Otherwise, it's too easy to accidentally miss that one of the recipients won't be able to read the message.

Also to be consistent with ProtonMail/crypto: https://github.com/ProtonMail/crypto/blob/1fa7f403fb9c3aa7b47443e1e3b39c80d6eedcb6/openpgp/write.go#L328-L333
I would just error when a key has no valid encryption key.

PR #61, opened 2020-07-16
Add EncryptRawMessageArmored helper function

hcmlinj:
To generates an armored PGP message given a rawdata and an armored public key.

### DIFF
--- a/helper/helper.go
+++ b/helper/helper.go
@@ -69,6 +69,36 @@ func EncryptMessageArmored(key, plaintext string) (ciphertext string, err error)
 	return ciphertext, nil
 }
 
+// EncryptRawMessageArmored generates an armored PGP message given a rawdata and
+// an armored public key.
+func EncryptRawMessageArmored(key string, data []byte) (string, error) {
+	publicKey, err := crypto.NewKeyFromArmored(key)
+
+	if err != nil {
+		return "", err
+	}
+
+	publicKeyRing, err := crypto.NewKeyRing(publicKey)
+
+	if err != nil {
+		return "", err
+	}
+
+	pgpMessage, err := publicKeyRing.Encrypt(crypto.NewPlainMessage(data), nil)
+
+	if err != nil {
+		return "", err
+	}
+
+	ciphertext, err := pgpMessage.GetArmored()
+
+	if err != nil {
+		return "", err
+	}
+
+	return ciphertext, nil
+}
+
 // EncryptSignMessageArmored generates an armored signed PGP message given a
 // plaintext and an armored public key a private key and its passphrase.
 func EncryptSignMessageArmored(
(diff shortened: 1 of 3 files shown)


helper/helper.go
```diff
@@ -69,6 +69,36 @@ func EncryptMessageArmored(key, plaintext string) (ciphertext string, err error)
 	return ciphertext, nil
 }
 
+// EncryptRawMessageArmored generates an armored PGP message given a rawdata and
+// an armored public key.
+func EncryptRawMessageArmored(key string, data []byte) (string, error) {
```
wussler, line 74:
```suggestion
func EncryptBinaryMessageArmored(key string, data []byte) (string, error) {
```

helper/helper_test.go
```diff
@@ -49,6 +49,30 @@ func TestArmoredTextMessageEncryption(t *testing.T) {
 	assert.Exactly(t, plaintext, decrypted)
 }
 
+func TestRawMessageArmoredEncryption(t *testing.T) {
```
wussler, line 52:
```suggestion
func TestArmoredBinaryMessageEncryption(t *testing.T) {
```

helper/helper_test.go
```diff
@@ -49,6 +49,30 @@ func TestArmoredTextMessageEncryption(t *testing.T) {
...
+	plaintext := "Secret message"
+
+	armored, err := EncryptRawMessageArmored(readTestFile("keyring_publicKey", false), []byte(plaintext))
```
wussler, line 55:
```suggestion
	armored, err := EncryptBinaryMessageArmored(readTestFile("keyring_publicKey", false), plainData)
```

helper/helper_test.go
```diff
@@ -49,6 +49,30 @@ func TestArmoredTextMessageEncryption(t *testing.T) {
 	assert.Exactly(t, plaintext, decrypted)
 }
 
+func TestRawMessageArmoredEncryption(t *testing.T) {
+	plaintext := "Secret message"
```
wussler, line 53:
```suggestion
	plainData := []byte("Secret message")
```

helper/helper.go
wussler, line 74: The function contains a lot of duplicated code with `EncryptMessageArmored`, could you create a common internal function that encrypts given the `PlainMessage` and use it in both?

helper/helper_test.go
```diff
@@ -49,6 +49,30 @@ func TestArmoredTextMessageEncryption(t *testing.T) {
...
+
+	armored, err := EncryptRawMessageArmored(readTestFile("keyring_publicKey", false), []byte(plaintext))
+
+	if err != nil {
+		t.Fatal("Expected no error when encrypting, got:", err)
+	}
+
+	assert.Exactly(t, true, crypto.IsPGPMessage(armored))
+
+	decrypted, err := DecryptMessageArmored(
```
wussler, line 63: Please edit `DecryptMessageArmored`, (like the encrypt function) to have a common internal function with two wrappers:
- `DecryptMessageArmored` with the same signature 
- `DecryptBinaryMessageArmored` returning `[]byte`.

If we expect bytes it makes more sense not to cast the data to a string

hcmlinj:
Hi @wussler ,
I've done the commit according to your suggestion.
Any deficiencies, please let me known.
I'm very happy can make contribution.

PR #62, opened 2020-07-16
Add session key size check

wussler:
Check session key size on encryption and decryption, throw an error if it does not match the expected size.

### DIFF
--- a/crypto/sessionkey.go
+++ b/crypto/sessionkey.go
@@ -104,7 +104,7 @@ func newSessionKeyFromEncrypted(ek *packet.EncryptedKey) (*SessionKey, error) {
 	}
 
 	if err := sk.checkSize(); err != nil {
-		return nil, errors.Wrap(err, "gopenpgp: unable to decode session key")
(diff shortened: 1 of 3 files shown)


crypto/sessionkey.go
```diff
@@ -98,12 +98,16 @@ func newSessionKeyFromEncrypted(ek *packet.EncryptedKey) (*SessionKey, error) {
...
 
-	symmetricKey := &SessionKey{
+	sk := &SessionKey{
 		Key:  ek.Key,
 		Algo: algo,
 	}
 
-	return symmetricKey, nil
+	if err := sk.checkSize(); err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: unable to decode session key")
```
twiss, line 107:
```suggestion
		return nil, errors.Wrap(err, "gopenpgp: unable to decrypt session key")
```
Since this function is only called from `DecryptSessionKey` and I think it doesn't really make sense in any other context

PR #64, opened 2020-07-21
Add check for updating server time in UpdateTime.

dimitar10:


crypto/time.go
```diff
@@ -7,7 +7,9 @@ import (
 
 // UpdateTime updates cached time.
 func UpdateTime(newTime int64) {
-	pgp.latestServerTime = newTime
+	if(newTime > pgp.latestServerTime) {
```
twiss, line 10:
```suggestion
	if (newTime > pgp.latestServerTime) {
```
The linter complains about this

crypto/time.go
```diff
@@ -7,7 +7,9 @@ import (
...
+		pgp.latestServerTime = newTime
+	}
 	pgp.latestClientTime = time.Now()
```
twiss, line 13: I would put this line inside the if as well, otherwise `getDiff()` calculation below will return some strange results.

PR #73, opened 2020-08-24
new helpers for drive decryption

marinthiercelin:
This PR adds new helper functions for the ProtonDrive codebase.

### DIFF
--- a/.golangci.yml
+++ b/.golangci.yml
@@ -27,3 +27,5 @@ linters:
     - testpackage       # Makes you use a separate _test package [fast: true, auto-fix: false]
     - wsl               # Whitespace Linter - Forces you to use empty lines! [fast: true, auto-fix: false]
     - gofumpt           # Enforce a stricter format than gofmt
+    - gci               # Enforce blank lines check
+    - nlreturn          # Enforce blank lines for return statements
--- a/helper/helper.go
+++ b/helper/helper.go
@@ -329,7 +329,7 @@ func decryptMessageArmored(privateKey string, passphrase []byte, ciphertext stri
 	return message, nil
 }
 
-func signDetachedArmored(privateKey string, passphrase []byte, message *crypto.PlainMessage) (signature signature, err error) {
+func signDetachedArmored(privateKey string, passphrase []byte, message *crypto.PlainMessage) (signature string, err error) {
 	privateKeyObj, err := crypto.NewKeyFromArmored(privateKey)
 
 	if err != nil {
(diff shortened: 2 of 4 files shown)


marinthiercelin:
@wussler  Here's the PR with the new helpers.
There is quite some code redundancy between different helper functions, maybe we can make it a bit more modular. 
What do you think ?

wussler:
Hi @marinthiercelin, I agree with your opinion: I'd make it a bit more modular, but I think you can use the internal function`decryptMessageArmored`, then just perform the verification in the helper `DecryptVerifyArmoredDetached`.

Where was `DecryptSplitMessage` useful? Didn't we talk about not using the split messages?
I'd add in the same PR the function `EncryptSignArmoredDetached` to have its sibling here :)

Thanks, Aron

marinthiercelin:
The DecryptSplitMessage is helping in the decryptBinary method of the Decryptor.
Since the binary encrypted files are split in a keypacket and datapacket, we need a split message to decrypt.
Until now it was done directly in the swift code, so I added a helper to make it cleaner, but we can leave it in the swift code if you prefer.
I agree about `EncryptSignArmoredDetached`, I'll add that

marinthiercelin:
I added the sibling functions for several helper functions. And added internal methods for more modularity.
That's quite a lot of rewriting so we should check I didn't break anything.
And as I told you we have an issue when doing EncryptSignDetachedArmored, as we have to return both the ciphertext and detached signature, along with the error, but we can't have several return arguments through gomobile.

marinthiercelin:
Also, It seems in all these helper functions we use the "Encrypt-and-Sign" which is not recommended 
(cf [this discussion](https://crypto.stackexchange.com/questions/5458/should-we-sign-then-encrypt-or-encrypt-then-sign))
As we provide a signature of the plaintext directly, it can compromise the confidentiality.
Maybe we could think about using encrypt-then-sign or sign-then-encrypt instead.

marinthiercelin:
Oops, seems like I rewrote existing functions. I removed the conflicts hopefully.

wussler:
LGTM, thanks :)
I'd simply add a test for each exported function, to test the "correct outcome", and the mobile wrapper to allow Swift/Java to use the multiple returned values with a custom structure (in `mobile.go`)

wussler:
Ah, don't worry about the failed test
```
panic: unsupported aead mode [recovered]
```
It's an upstream bug in x/crypto, we're working on a fix

wussler (approved):
LGTM, maybe just a nitpick about the naming of `EncryptSignArmoredDetachedMobile` :)

Edit: and the CI is crying

wussler (approved):

PR #74, opened 2020-08-26
break cycle on first non-session-key packet

chesnokovilya:
break cycle packet.next on first non-session-key packet in getEncryptionKeyIDs

### DIFF
--- a/crypto/message.go
+++ b/crypto/message.go
@@ -239,15 +239,11 @@ Loop:
 		case *packet.EncryptedKey:
 			encryptedKey = p
 			ids = append(ids, encryptedKey.KeyId)
-		case *packet.SymmetricallyEncrypted:
+		case *packet.SymmetricallyEncrypted,
+			*packet.AEADEncrypted,
+			*packet.Compressed,
+			*packet.LiteralData:
 			break Loop
-		case *packet.AEADEncrypted:
-			break Loop
-		case *packet.Compressed:
-			break Loop
-		case *packet.LiteralData:
-			break Loop
-		default:
 		}
 	}
 	if len(ids) > 0 {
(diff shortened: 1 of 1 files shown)


wussler (approved):

wussler (approved):

crypto/message.go
```diff
@@ -228,14 +228,25 @@ func (msg *PGPMessage) getEncryptionKeyIDs() ([]uint64, bool) {
...
+			encryptedKey = p
+			ids = append(ids, encryptedKey.KeyId)
+		case *packet.SymmetricallyEncrypted:
+			break
+		case *packet.AEADEncrypted:
+			break
+		case *packet.Compressed:
+			break
+		case *packet.LiteralData:
+			break
```
twiss, line 248:
```suggestion
		case *packet.SymmetricallyEncrypted:
		case *packet.AEADEncrypted:
		case *packet.Compressed:
		case *packet.LiteralData:
			break
```

chesnokovilya: Golang does not have default fallthrough:  In a case or default clause, the last non-empty statement may be a (possibly labeled) "fallthrough" statement to indicate that control should flow from the end of this clause to the first statement of the next clause. Otherwise control flows to the end of the "switch" statement. A "fallthrough" statement may appear as the last statement of all but the last clause of an expression switch.

twiss: Ah yeah, sorry. But maybe you can do this:
```
		case *packet.SymmetricallyEncrypted, *packet.AEADEncrypted, *packet.Compressed, *packet.LiteralData:
			break
```

chesnokovilya: yeah, through maybe using simpler case condition is making it easier to understand at first glance.

twiss: Yeah, but I think repeating the case body 4 times is also a bit ugly, especially if we're gonna have this code in multiple places. If you want, you could split up the four case values above across 4 lines, to make it easier to read, though :man_shrugging:

PR #75, opened 2020-08-31
use aes256 for password encryption

chesnokovilya:


crypto/message_test.go
```diff
@@ -18,6 +19,25 @@ func TestTextMessageEncryptionWithPassword(t *testing.T) {
 	if err != nil {
 		t.Fatal("Expected no error when encrypting, got:", err)
 	}
+	packets := packet.NewReader(bytes.NewReader(encrypted.GetBinary()))
+	var foundSk bool
+	for {
+		var p packet.Packet
+		var errEOF error
+		if p, errEOF = packets.Next(); errEOF == io.EOF {
+			errEOF = nil
```
twiss, line 28: The linter is complaining about this assignment. Also, if `packets.Next()` returns an error other than EOF, I guess it's not handled right now

twiss (approved):
Looks good to me, other than nitpick below :+1:

crypto/message_test.go
chesnokovilya: Fixed linter complain. I guess here we are not expecting other errors here. If they appear it would throw later in `if !skFound`.

PR #78, opened 2020-09-11
New helpers for ios drive

marinthiercelin:


helper/helper.go
```diff
@@ -418,3 +449,23 @@ func decryptAttachment(
...
+	publicKeyObj, err := crypto.NewKeyFromArmored(publicKey)
+
+	if publicKeyObj.IsPrivate() {
+		publicKeyObj, err = publicKeyObj.ToPublic()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if err != nil {
```
wussler, line 465: Check the error first, `publicKeyObj` might be nil and panic

marinthiercelin: Right, Thanks

wussler (approved):

PR #83, opened 2020-10-08
Added helpers with encrypted signatures and unarmored binary ciphertexts

marinthiercelin:


helper/helper.go
```diff
@@ -347,48 +355,69 @@ func DecryptSessionKey(
...
+
 	if err != nil {
-		return "", errors.Wrap(err, "gopenpgp: unable to armor message")
+		return nil, err
+	}
+
+	ciphertext, err := publicKeyRing.Encrypt(message, nil)
+
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: unable to encrypt message")
```
twiss, line 392: This is a bit nitpicky but you added newlines before the `if err != nil` in a bunch of places in this PR, could you keep the error handling next to the line that it's.. "handling"? ^.^

twiss (approved):

PR #89, opened 2020-10-15
Migrate CI to actions, build artifacts

wussler:
- Use Actions instead of Travis
- Build for iOS, mac OS, and Android
- Make artifacts available

Co-authored by: @marinthiercelin


wussler:
I promise to squash when merging :sweat_smile:

wussler:
And here are the test artifacts:
- https://github.com/wussler/gopenpgp/actions/runs/308806560
- https://github.com/wussler/gopenpgp/actions/runs/308806551

marinthiercelin (commented):
In android.yml I think you checkout twice.
I am not sure the build.sh will work directly from inside the directory of gopenpgp

wussler:
@marinthiercelin Yes, it works directly there, but I'm not 100% sure whether on the current (checked-out) version.

You're right, on android we check-out twice, I'll fix that

marinthiercelin:
Can you put the Crypto.xcframework in the artifacts as well? it combines the frameworks for all appleplatforms
Btw, maybe we can modify the script to call it Gopenpgp.framework instead

wussler:
@marinthiercelin I like the name gopenpgp too, renamed and added :+1:

marinthiercelin (commented):
You also need to modify $BUILD_NAME in build.sh, otherwise it won't genereate with the right name

marinthiercelin:
And why do you upload a Gopenpgp.framework instead of Gopenpgp.xcframework ?
I should be xcframework

marinthiercelin (commented):
Why do you delete the remove_dir bash function ?
I use it in the build function. So in its current state, the script is goind to fail

wussler:
Now everything works, just not renamed :)

wussler:
It's now renamed and ready for merge :)

marinthiercelin (approved):

PR #95, opened 2020-11-04
Dummy key.Check()

larabr:
Key param checks are implemented by https://github.com/ProtonMail/crypto/pull/59 and performed as part of key parsing, hence `key.Check()` is redundant and can be converted to a dummy function.

This PR can me merged once we update the ProtonMail/crypto dependency to include the relevant changes.


wussler:
Integrated in https://github.com/ProtonMail/gopenpgp/pull/91/files

PR #99, opened 2020-11-13
Allow multiple keypackets in session key decryption

wussler:


crypto/keyring_session.go
```diff
@@ -9,28 +9,53 @@ import (
...
+			*packet.LiteralData:
+			break Loop
+
+		default:
+			continue Loop
+		}
+	}
+
+	if !hasPacket {
+		return nil, errors.Wrap(err, "gopenpgp: error in reading packets")
```
twiss, line 58:
```suggestion
		return nil, errors.Wrap(err, "gopenpgp: couldn't find a session key packet that could be decrypted")
```

twiss (approved):
LGTM :+1: One tiny suggestion below, to make the error message slightly more informative

PR #102, opened 2020-11-18
Drop regex for canonicalization

wussler:
Change canonicalization to `strings.TrimRight` instead of regex

### DIFF
--- a/crypto/message.go
+++ b/crypto/message.go
@@ -88,7 +88,7 @@ func NewPlainMessageFromFile(data []byte, filename string, time uint32) *PlainMe
 // ready for encryption, signature, or verification from an unencrypted string.
 func NewPlainMessageFromString(text string) *PlainMessage {
 	return &PlainMessage{
-		Data:     []byte(strings.ReplaceAll(strings.ReplaceAll(text, "\r\n", "\n"), "\n", "\r\n")),
+		Data:     []byte(internal.CanonicalizeAndTrim(text)),
 		TextType: true,
 		Filename: "",
 		Time:     uint32(GetUnixTime()),
--- a/helper/cleartext_test.go
+++ b/helper/cleartext_test.go
@@ -2,11 +2,11 @@ package helper
 
 import (
 	"regexp"
-	"strings"
 	"testing"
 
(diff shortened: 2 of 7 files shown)


twiss (approved):

helper/cleartext_test.go
```diff
@@ -45,11 +45,5 @@ func TestSignClearText(t *testing.T) {
 	if err != nil {
 		t.Fatal("Cannot parse message:", err)
 	}
-	assert.Exactly(t, canonicalizeAndTrim(inputPlainText), string(clearTextMessage.GetBinary()))
-}
-
-func canonicalizeAndTrim(text string) string {
```
henrybear327, line 51: I think the refactored code change is not the same as the original regex. 

In the original implementation, only the trailing space of the entire string input is trimmed, and then all the newlines are globally replaced with \r\n.

But, in the refactored implementation, spaces are trimmed on every line.

crypto/message.go
```diff
@@ -88,7 +88,7 @@ func NewPlainMessageFromFile(data []byte, filename string, time uint32) *PlainMe
 // ready for encryption, signature, or verification from an unencrypted string.
 func NewPlainMessageFromString(text string) *PlainMessage {
 	return &PlainMessage{
-		Data:     []byte(strings.ReplaceAll(strings.ReplaceAll(text, "\r\n", "\n"), "\n", "\r\n")),
+		Data:     []byte(internal.CanonicalizeAndTrim(text)),
```
zhj4478, line 91: we shouldn't trim all inputs. we should only trim when we need a trim. I think only clear-text signatures need to trim.

helper/cleartext_test.go
wussler: This is intended behaviour according to RFC 4880 paragraphs 5.2.1 and 7.1

crypto/message.go
wussler: On the BE we convert some instances of signed messages to cleartext according to the preferences, so this is as of now needed for all messages. Please refer to https://github.com/ProtonMail/gopenpgp/pull/136

PR #104, opened 2020-12-01
Add key generation offset

wussler:
Add function `SetKeyGenerationOffset` to add an offset in key generation time and prevent not-yet-valid keys


marinthiercelin (approved):
Apart from possible overflows when adding the time with extapolate and the offset, LGTM.

PR #106, opened 2020-12-08
Update x/crypto fork to github.com/ProtonMail/go-crypto

twiss:
This way, users of gopenpgp don't need to add a `replace` to their go.mod anymore.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,13 +3,11 @@ module github.com/ProtonMail/gopenpgp/v2
 go 1.15
 
 require (
+	github.com/ProtonMail/go-crypto v0.0.0-20201208171014-cdb7591792e2
 	github.com/ProtonMail/go-mime v0.0.0-20190923161245-9b5a4261663a
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.4.0
-	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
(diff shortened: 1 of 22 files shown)


go.mod
```diff
@@ -3,13 +3,11 @@ module github.com/ProtonMail/gopenpgp/v2
...
 	github.com/ProtonMail/go-mime v0.0.0-20190923161245-9b5a4261663a
 	github.com/pkg/errors v0.8.1
 	github.com/stretchr/testify v1.4.0
-	golang.org/x/crypto v0.0.0-20191011191535-87dc89f01550
 	golang.org/x/mobile v0.0.0-20200801112145-973feb4309de
 )
 
-replace golang.org/x/crypto => github.com/ProtonMail/crypto v0.0.0-20201112115411-41db4ea0dd1c
-
 replace golang.org/x/mobile => github.com/marinthiercelin/mobile v0.0.0-20201127122539-61ba718dc1d1
```
marinthiercelin, line 13: We can also change the go mobile replacement to use ProtonMail/go-mobile instead of my fork

marinthiercelin: If we do, and we make it point to the master branch, then we need to make sure the github build actions use golang 15.6 and not golang 15.5.

twiss: Maybe you can do that in a separate PR afterwards?

marinthiercelin: Allright

marinthiercelin (approved):

PR #107, opened 2020-12-08
Change gomobile fork to use the github.com/ProtonMail/go-mobile fork

marinthiercelin:
We've moved the go-mobile fork to the PM github space.
We update the github actions to use golang 1.15.6, as 15.5 had a bug that made gomobile fail on ios.
I also changed build.sh to have dist as the build dir, as it was originally used for this intent


twiss (approved):
:+1:
(Could you remove the github.com/marinthiercelin/mobile from go.sum? If it's no longer needed)

marinthiercelin:
> +1
> (Could you remove the github.com/marinthiercelin/mobile from go.sum? If it's no longer needed)

PR #111, opened 2021-01-13
Draft: Decryption with io.Reader

marinthiercelin:
Introduce two KeyRing methods that does stream decryption, taking a io.Reader for either an armored or unarmored pgpMessage.
It returns a io.Reader for the decrypted data.


marinthiercelin:
Currently the methods completely ignore the verification of the decrypted data.
Is that acceptable.
I also need to add tests and modify the changelog

PR #117, opened 2021-03-08
Fix various keys issues

wussler:
- Armouring headers for public or private keys must be different
- Session key decoding on invalid keys returns now the right error


crypto/keyring_session.go
```diff
@@ -55,14 +55,14 @@ Loop:
...
+		return nil, errors.Wrap(err, "gopenpgp: couldn't find a session key packet")
 	}
 
 	if decryptErr != nil {
 		return nil, errors.Wrap(decryptErr, "gopenpgp: error in decrypting")
 	}
 
-	if ek == nil {
+	if ek == nil || ek.Key == nil {
 		return nil, errors.New("gopenpgp: unable to decrypt session key")
```
wussler, line 66:
```suggestion
		return nil, errors.New("gopenpgp: unable to decrypt session key: no valid decryption key")
```

twiss (approved):

PR #120, opened 2021-03-26
New attachment processor

marinthiercelin:
We introduce a new attachment processor where the caller has to pre-allocate a buffer for the data packet.
This can avoid memory management issues in mobile applications.


crypto/attachment.go
```diff
@@ -13,6 +15,178 @@ import (
...
+	}
+	if err := ap.plaintextWriter.Close(); err != nil {
+		return errors.Wrap(err, "gopengpp: unable to close plaintext writer")
+	}
+	if err := ap.ciphertextWriter.Close(); err != nil {
+		return errors.Wrap(err, "gopengpp: unable to close the dataPacket writer")
+	}
+	ap.done.Wait()
+	if ap.err != nil {
+		fmt.Printf("Error while finishing %v\n", ap.err)
```
wussler, line 64: Is this intended or a leftover from debugging?

crypto/attachment.go
```diff
@@ -13,6 +15,178 @@ import (
 	"github.com/pkg/errors"
 )
 
+// AttachmentProcessor2 keeps track of the progress of encrypting an attachment
+// (optimized for encrypting large files).
+// With this processor, the caller has to first allocate
+// a buffer large enough to hold the whole data packet.
+type AttachmentProcessor2 struct {
```
wussler, line 22:
```suggestion
type ManualAttachmentProcessor struct {
```

Or something that recalls the manual/aggressive memory management :)

I'd also put the new types at the beginning of the file or in a separate file, followed then by the methods as we do in the rest of the lib

crypto/attachment.go
```diff
@@ -13,6 +15,178 @@ import (
...
+	}
+
+	// forces the gc to be called often
+	debug.SetGCPercent(10)
+
+	attachmentProc := &AttachmentProcessor2{}
+
+	// hints for the encrypted file
+	isBinary := true
+	modTime := uint32(GetUnixTime())
```
wussler, line 90: Why uint32 to then cast to an int64?

crypto/attachment.go
marinthiercelin: It's from debugging yes.

crypto/attachment.go
marinthiercelin: I think a separate file would make sense.
would `attachment_manual.go` work with you ?

crypto/attachment.go
marinthiercelin: I don't know, I just copied the way it's done in the old processor

wussler: I suspected, probably we should change this. @twiss is there any reason I'm missing here?

crypto/attachment.go
wussler: Yup, good for me :+1:

crypto/attachment.go
twiss: Well, timestamps in OpenPGP are indeed uint32 (:sweat_smile:).
The casting from a time to a number and back seems unnecessary indeed though, this can be left to go-crypto.

marinthiercelin: Ok I'll remove the unnecessary casting then

twiss (approved):

PR #126, opened 2021-04-05
feat: allow using a custom packet.Config

djgilcrease:
related-to: https://github.com/ProtonMail/go-crypto/pull/79
related-to: https://github.com/goreleaser/nfpm/pull/315

### DIFF
--- a/crypto/keyring.go
+++ b/crypto/keyring.go
@@ -16,6 +16,9 @@ type KeyRing struct {
 
 	// FirstKeyID as obtained from API to match salt
 	FirstKeyID string
+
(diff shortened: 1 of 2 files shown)


twiss:
Hey :wave: Rather than passing a go-crypto config to gopenpgp, I would prefer if gopenpgp's API remain self-contained. If absolutely necessary, we could add a separate config for gopenpgp. However, selecting the signing subkey seems like a rather low-level concern, that users of gopenpgp typically shouldn't be concerned with. If at all possible I would try one of the other approaches I mentioned in the linked go-crypto PR.

PR #127, opened 2021-04-06
WIP: feat: allow setting a custom signing hash as well as specifying the signing key id to use.

djgilcrease:
requires: https://github.com/ProtonMail/go-crypto/pull/79

* Tried to keep the API self-contained this time and am not exposing the whole packet config object.
* Added `WithSigningHash(hash crypto.Hash)` to be able to set the hash used to sign things with `SignDetached`
* Added `SignDetachedWithKeyID(message *PlainMessage, keyId uint64) (*PGPSignature, error)` that sets the `packet.Config.SigningKeyId`
* Changed `SignDetached(message *PlainMessage) (*PGPSignature, error)` to just call `SignDetachedWithKeyID(message *PlainMessage, keyId uint64)` with the keyId as `0`
* Added `SignDetachedEncryptedWithKeyID(message *PlainMessage, encryptionKeyRing *KeyRing, keyId uint64)`


wussler:
Hi @djgilcrease, thank you for the PR. Unfortunately I think this functionality is not suitable for gopenpgp's API.
In order to force encryption with a specific key ID I'd recommend using go-crypto directly.

Our objective is to keep gopenpgp's API "batteries included", providing sensible defaults to the security options, and not allowing users to force encryption with keys that might lead to bad packets or errors difficult to debug.
@twiss confirmed that there is intention to merge the upstream PR to provide this functionality there.

PR #128, opened 2021-04-26
Add methods for embedded signatures using session keys

wussler:
Add:
- SessionKey methods to encrypt/decrypt and simultaneously sign/verify with an asymmetric key (embedded signature)
```go
(sk *SessionKey) EncryptAndSign(message *PlainMessage, signKeyRing *KeyRing) ([]byte, error)
(sk *SessionKey) DecryptAndVerify(dataPacket []byte, verifyKeyRing *KeyRing, verifyTime int64) (*PlainMessage, error)
```
- The mobile helper `DecryptSessionKeyExplicitVerify` to allow using session key decryption + verification operations via gomobile

### DIFF
--- a/crypto/sessionkey.go
+++ b/crypto/sessionkey.go
@@ -124,7 +124,30 @@ func (sk *SessionKey) Encrypt(message *PlainMessage) ([]byte, error) {
 		DefaultCipher: dc,
 	}
 
-	return encryptWithSessionKey(message, sk, config)
+	return encryptWithSessionKey(message, sk, nil, config)
+}
+
+// EncryptAndSign encrypts a PlainMessage to PGPMessage with a SessionKey and signs it with a Private key.
+// * message : The plain data as a PlainMessage.
+// * signKeyRing: The KeyRing to sign the message
+// * output  : The encrypted data as PGPMessage.
+func (sk *SessionKey) EncryptAndSign(message *PlainMessage, signKeyRing *KeyRing) ([]byte, error) {
+	dc, err := sk.GetCipherFunc()
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: unable to encrypt with session key")
+	}
+
+	config := &packet.Config{
+		Time:          getTimeGenerator(),
+		DefaultCipher: dc,
+	}
+
+	signEntity, err := signKeyRing.getSigningEntity()
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: unable to sign")
+	}
+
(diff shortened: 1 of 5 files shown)


crypto/sessionkey.go
```diff
@@ -162,37 +185,70 @@ func encryptWithSessionKey(message *PlainMessage, sk *SessionKey, config *packet
...
+		hints := &openpgp.FileHints{
+			IsBinary: message.IsBinary(),
+			FileName: message.Filename,
+			ModTime:  message.getFormattedTime(),
+		}
 
-	if err != nil {
-		return nil, errors.Wrap(err, "gopenpgp: unable to serialize")
-	}
+		signWriter, err = openpgp.Sign(encryptWriter, signEntity, hints, config)
```
twiss, line 195: I'm not a super big fan of using the top-level `openpgp.Sign` as a building block, though I guess there's not an obvious better option. It has subtle implications, such as it uses the signer's hash algorithm preferences rather than the recipient's, and if we ever implement intended recipient fingerprints, this would break that. I guess it's fine for now but we should keep this in mind in the future.

Also, since this is apparently the first time we use `openpgp.Sign`, that function has been left behind a bit, e.g. it falls back to RIPEMD [here](https://github.com/ProtonMail/go-crypto/blob/da7cae9ac5f28d6ee1d856d8a079f10c8c015f47/openpgp/write.go#L455) - we should at least fix that to fall back to SHA256, akin to [this](https://github.com/ProtonMail/go-crypto/blob/da7cae9ac5f28d6ee1d856d8a079f10c8c015f47/openpgp/write.go#L349). Also, there's a theoretical case where there's no intersection between the sender's preferred hashes and the candidate hashes, in which case I think we would panic, so we should add an error akin to [this one](https://github.com/ProtonMail/go-crypto/blob/da7cae9ac5f28d6ee1d856d8a079f10c8c015f47/openpgp/write.go#L391-L393). Could you make a PR for that, perhaps?

Other than that, it looks good to me :+1:

wussler: I see, thanks for the insight. I agree that using `Sign()` directly is not ideal, but I couldn't really find anything better. When we'll have intended recipients I guess we'll need another function (to replace/integrate `Sign` as well).

I'll make a PR for the defaults as you suggested :)

twiss (approved):

wussler:
Note: go-crypto will be updated to contain https://github.com/ProtonMail/go-crypto/commit/80b97ac9f3947fbe5812240d3ddf7e81be0558b7 in a later PR

PR #131, opened 2021-05-11
Add a streaming api to KeyRing and SessionKey

marinthiercelin:
This PR adds new methods to KeyRing and SessionKey to allow encryption/decryption and signing/verification using io.Reader and io.Writer directly instead of having to load the data in memory beforehand.
It also adds helpers for mobile apps to be able to implement Readers and Writers to be used with these methods

### DIFF
--- a/crypto/keyring_message.go
+++ b/crypto/keyring_message.go
@@ -119,7 +119,7 @@ func (keyRing *KeyRing) VerifyDetachedEncrypted(message *PlainMessage, encrypted
 
 // ------ INTERNAL FUNCTIONS -------
 
-// Core for encryption+signature functions.
+// Core for encryption+signature (non-streaming) functions.
 func asymmetricEncrypt(
 	plainMessage *PlainMessage,
 	publicKey, privateKey *KeyRing,
@@ -153,6 +153,7 @@ func asymmetricEncrypt(
 	return outBuf.Bytes(), nil
 }
 
+// Core for encryption+signature (all) functions.
 func asymmetricEncryptStream(
 	hints *openpgp.FileHints,
 	keyPacketWriter io.Writer,
@@ -181,7 +182,7 @@ func asymmetricEncryptStream(
 	return encryptWriter, nil
 }
 
-// Core for decryption+verification functions.
+// Core for decryption+verification (non streaming) functions.
 func asymmetricDecrypt(
 	encryptedIO io.Reader, privateKey *KeyRing, verifyKey *KeyRing, verifyTime int64,
 ) (message *PlainMessage, err error) {
@@ -212,7 +213,7 @@ func asymmetricDecrypt(
 	}, err
 }
 
-// Core for decryption+verification functions.
+// Core for decryption+verification (all) functions.
 func asymmetricDecryptStream(
 	encryptedIO io.Reader, privateKey *KeyRing, verifyKey *KeyRing,
 ) (messageDetails *openpgp.MessageDetails, err error) {
--- a/crypto/keyring_message.go
+++ b/crypto/keyring_message.go
@@ -155,8 +155,8 @@ func asymmetricEncrypt(
 
 func asymmetricEncryptStream(
 	hints *openpgp.FileHints,
-	dataPacketWriter io.Writer,
 	keyPacketWriter io.Writer,
+	dataPacketWriter io.Writer,
 	publicKey, privateKey *KeyRing,
 	config *packet.Config,
 ) (encryptWriter io.WriteCloser, err error) {
--- a/crypto/keyring_message.go
+++ b/crypto/keyring_message.go
@@ -127,16 +127,7 @@ func asymmetricEncrypt(
 ) ([]byte, error) {
 	var outBuf bytes.Buffer
 	var encryptWriter io.WriteCloser
-	var signEntity *openpgp.Entity
 	var err error
-
-	if privateKey != nil && len(privateKey.entities) > 0 {
-		var err error
-		signEntity, err = privateKey.getSigningEntity()
-		if err != nil {
-			return nil, err
-		}
-	}
 
 	hints := &openpgp.FileHints{
 		IsBinary: plainMessage.IsBinary(),
@@ -144,11 +135,7 @@ func asymmetricEncrypt(
 		ModTime:  plainMessage.getFormattedTime(),
 	}
 
-	if plainMessage.IsBinary() {
-		encryptWriter, err = openpgp.Encrypt(&outBuf, publicKey.entities, signEntity, hints, config)
-	} else {
-		encryptWriter, err = openpgp.EncryptText(&outBuf, publicKey.entities, signEntity, hints, config)
-	}
+	encryptWriter, err = asymmetricEncryptStream(hints, &outBuf, publicKey, privateKey, config)
 	if err != nil {
 		return nil, errors.Wrap(err, "gopenpgp: error in encrypting asymmetrically")
 	}
@@ -166,24 +153,42 @@ func asymmetricEncrypt(
 	return outBuf.Bytes(), nil
 }
 
+func asymmetricEncryptStream(
+	hints *openpgp.FileHints,
+	outWriter io.Writer,
+	publicKey, privateKey *KeyRing,
+	config *packet.Config,
+) (encryptWriter io.WriteCloser, err error) {
+	var signEntity *openpgp.Entity
+
+	if privateKey != nil && len(privateKey.entities) > 0 {
+		var err error
+		signEntity, err = privateKey.getSigningEntity()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	if hints.IsBinary {
+		encryptWriter, err = openpgp.Encrypt(outWriter, publicKey.entities, signEntity, hints, config)
+	} else {
+		encryptWriter, err = openpgp.EncryptText(outWriter, publicKey.entities, signEntity, hints, config)
+	}
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: error in encrypting asymmetrically")
+	}
+	return encryptWriter, nil
+}
+
 // Core for decryption+verification functions.
 func asymmetricDecrypt(
 	encryptedIO io.Reader, privateKey *KeyRing, verifyKey *KeyRing, verifyTime int64,
 ) (message *PlainMessage, err error) {
-	privKeyEntries := privateKey.entities
-	var additionalEntries openpgp.EntityList
-
-	if verifyKey != nil {
-		additionalEntries = verifyKey.entities
-	}
-
-	if additionalEntries != nil {
-		privKeyEntries = append(privKeyEntries, additionalEntries...)
-	}
-
-	config := &packet.Config{Time: getTimeGenerator()}
-
-	messageDetails, err := openpgp.ReadMessage(encryptedIO, privKeyEntries, nil, config)
+	messageDetails, err := asymmetricDecryptStream(
+		encryptedIO,
+		privateKey,
+		verifyKey,
+	)
 	if err != nil {
 		return nil, errors.Wrap(err, "gopenpgp: error in reading message")
 	}
@@ -204,4 +209,28 @@ func asymmetricDecrypt(
 		Filename: messageDetails.LiteralData.FileName,
 		Time:     messageDetails.LiteralData.Time,
 	}, err
+}
+
+// Core for decryption+verification functions.
+func asymmetricDecryptStream(
+	encryptedIO io.Reader, privateKey *KeyRing, verifyKey *KeyRing,
+) (messageDetails *openpgp.MessageDetails, err error) {
+	privKeyEntries := privateKey.entities
+	var additionalEntries openpgp.EntityList
+
+	if verifyKey != nil {
+		additionalEntries = verifyKey.entities
+	}
+
+	if additionalEntries != nil {
+		privKeyEntries = append(privKeyEntries, additionalEntries...)
+	}
+
+	config := &packet.Config{Time: getTimeGenerator()}
+
+	messageDetails, err = openpgp.ReadMessage(encryptedIO, privKeyEntries, nil, config)
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: error in reading message")
+	}
+	return messageDetails, err
 }
--- a/crypto/keyring_message.go
+++ b/crypto/keyring_message.go
@@ -137,7 +137,7 @@ func asymmetricEncrypt(
 
 	encryptWriter, err = asymmetricEncryptStream(hints, &outBuf, &outBuf, publicKey, privateKey, config)
 	if err != nil {
-		return nil, errors.Wrap(err, "gopenpgp: error in encrypting asymmetrically")
+		return nil, err
 	}
 
 	_, err = encryptWriter.Write(plainMessage.GetBinary())
@@ -191,7 +191,7 @@ func asymmetricDecrypt(
 		verifyKey,
 	)
 	if err != nil {
-		return nil, errors.Wrap(err, "gopenpgp: error in reading message")
+		return nil, err
 	}
 
 	body, err := ioutil.ReadAll(messageDetails.UnverifiedBody)
--- a/crypto/keyring_message.go
+++ b/crypto/keyring_message.go
@@ -135,7 +135,7 @@ func asymmetricEncrypt(
 		ModTime:  plainMessage.getFormattedTime(),
 	}
 
-	encryptWriter, err = asymmetricEncryptStream(hints, &outBuf, publicKey, privateKey, config)
+	encryptWriter, err = asymmetricEncryptStream(hints, &outBuf, &outBuf, publicKey, privateKey, config)
 	if err != nil {
 		return nil, errors.Wrap(err, "gopenpgp: error in encrypting asymmetrically")
 	}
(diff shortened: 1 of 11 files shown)


crypto/sessionkey.go
```diff
@@ -171,65 +172,85 @@ func (sk *SessionKey) EncryptWithCompression(message *PlainMessage) ([]byte, err
...
+		encBuf,
+		sk,
+		signEntity,
+		config,
+	)
 	if err != nil {
-		return nil, errors.Wrap(err, "gopenpgp: unable to encrypt")
+		return nil, err
+	}
+	if signWriter != nil {
```
wussler, line 188: I'd rather check if `signEntity` is not nil, rather than relying on `signWriter`

crypto/sessionkey.go
```diff
@@ -273,45 +318,24 @@ func (sk *SessionKey) DecryptAndVerify(dataPacket []byte, verifyKeyRing *KeyRing
...
-		err = verifyDetailsSignature(md, verifyKeyRing)
-	}
-
-	return &PlainMessage{
-		Data:     messageBuf.Bytes(),
-		TextType: !md.LiteralData.IsBinary,
-		Filename: md.LiteralData.FileName,
-		Time:     md.LiteralData.Time,
-	}, err
+	return md, err
```
wussler, line 338:
```suggestion
	return md, nil
```

I like ensuring here the error's `nil` and we wrap the error

crypto/sessionkey_streaming.go
```diff
@@ -0,0 +1,97 @@
...
+}
+
+// DecryptStream is used to decrypt a data packet as a Reader.
+// It takes a reader for the data packet
+// and returns a PlainMessageReader for the plaintext data.
+// If verifyKeyRing is not nil, PlainMessageReader.VerifySignature() will
+// verify the embedded signature with the given key ring and verification time.
+func (sk *SessionKey) DecryptStream(
+	dataPacketReader Reader,
+	verifyKeyRing *KeyRing, verifyTime int64,
```
wussler, line 80: Either all on one line, or one line per type

helper/mobile_stream.go
```diff
@@ -0,0 +1,157 @@
...
+
+import (
+	"crypto/sha256"
+	"errors"
+	"hash"
+	"io"
+	"runtime"
+
+	"github.com/ProtonMail/gopenpgp/v2/crypto"
+	errorsWrap "github.com/pkg/errors"
```
wussler, line 11: You can replace all `errors` with this package, no need for the double import

crypto/keyring_streaming_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"io"
+	"testing"
+
+	"github.com/pkg/errors"
+)
+
+const testFilename = "filename.txt"
+
+func TestKeyRing_EncryptStream(t *testing.T) {
+	keyRingPrivate, err := keyRingTestPrivate.Copy()
```
wussler, line 14: Why copying? (Same for the public ones)

crypto/keyring_streaming_test.go
```diff
@@ -0,0 +1,228 @@
...
+	"bytes"
+	"io"
+	"testing"
+
+	"github.com/pkg/errors"
+)
+
+const testFilename = "filename.txt"
+
+func TestKeyRing_EncryptStream(t *testing.T) {
```
wussler, line 13: I'd also test reading the encrypted data and decrypting with the "normal" functions to prevent any kind of regression (even though I know they're now implemented using the streaming API under the hood)

crypto/keyring_streaming_test.go
```diff
@@ -0,0 +1,228 @@
...
+	}
+	if filename != decryptedReader.GetFilename() {
+		t.Fatalf("Expected filename to be %s got %s", filename, decryptedReader.GetFilename())
+	}
+	if modTime != decryptedReader.GetModificationTime() {
+		t.Fatalf("Expected modification time to be %d got %d", modTime, decryptedReader.GetModificationTime())
+	}
+}
+
+func TestKeyRing_EncryptSplitStream(t *testing.T) {
```
wussler, line 93: I'd also test reading the encrypted data packet and key packet, join them then try decrypting with the "normal" functions

crypto/sessionkey_streaming_test.go
```diff
@@ -0,0 +1,89 @@
...
+
+import (
+	"bytes"
+	"io"
+	"testing"
+
+	"github.com/pkg/errors"
+)
+
+func TestSessionKey_EncryptStream(t *testing.T) {
```
wussler, line 11: Same here, add some test that ensure compatibility with non-streaming API

wussler (commented):
Thanks @marinthiercelin, Great PR! Adds in a clean way a lot of useful functionality, seems pretty complete and with mobile helpers.

I've left a few comments, the most pressing one on the design level is whether we want to unify the "hints" management with a message, to decide with @twiss.

The changelog is missing, but I'd do that after rebasing on the latest master.

crypto/keyring_message.go
```diff
@@ -205,3 +211,27 @@ func asymmetricDecrypt(
 		Time:     messageDetails.LiteralData.Time,
 	}, err
 }
+
+// Core for decryption+verification functions.
```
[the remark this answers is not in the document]
marinthiercelin: It's not only for streaming functions though, it's also used by regular methods (to avoid code duplication).

crypto/keyring_message.go
```diff
@@ -166,26 +153,45 @@ func asymmetricEncrypt(
 	return outBuf.Bytes(), nil
 }
 
-// Core for decryption+verification functions.
-func asymmetricDecrypt(
-	encryptedIO io.Reader, privateKey *KeyRing, verifyKey *KeyRing, verifyTime int64,
-) (message *PlainMessage, err error) {
-	privKeyEntries := privateKey.entities
-	var additionalEntries openpgp.EntityList
+func asymmetricEncryptStream(
```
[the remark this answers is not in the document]
marinthiercelin: Same as for decryption, non-streaming methods are using it too.

crypto/keyring_streaming.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// EncryptStream is used to encrypt data as a Writer.
+// It takes a writer for the encrypted data and returns a writer for the plaintext data
+// If signKeyRing is not nil, it is used to do an embedded signature.
+func (keyRing *KeyRing) EncryptStream(
+	pgpMessageWriter Writer,
+	isBinary bool,
+	filename string,
+	modTime int64,
```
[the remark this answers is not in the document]
marinthiercelin: Yes I agree that it's a bit ugly to have all these hints one by one. Should we replace it with a struct ? I think calling it streaming message would be confusing though, as it wouldn't contain any data or reference to the data

crypto/keyring_streaming.go
```diff
@@ -0,0 +1,275 @@
...
+
+// DecryptStream is used to decrypt a pgp message as a Reader.
+// It takes a reader for the message data
+// and returns a PlainMessageReader for the plaintext data.
+// If verifyKeyRing is not nil, PlainMessageReader.VerifySignature() will
+// verify the embedded signature with the given key ring and verification time.
+func (keyRing *KeyRing) DecryptStream(
+	message Reader,
+	verifyKeyRing *KeyRing,
+	verifyTime int64,
```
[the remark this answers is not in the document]
marinthiercelin: For the verify time we could, but for the verification key ring, go-crypto/openpgp needs it before starting the decryption. At least I haven't found a way to give it afterward. 
So I kept them both at the start, because it keeps consistency of having both verify keyring and verify time together.

crypto/keyring_streaming.go
```diff
@@ -0,0 +1,275 @@
...
+}
+
+// VerifyDetachedStream verifies a message reader with a detached PGPSignature
+// and returns a SignatureVerificationError if fails.
+func (keyRing *KeyRing) VerifyDetachedStream(
+	message Reader,
+	signature *PGPSignature,
+	verifyTime int64,
+) error {
+	return verifySignature(
```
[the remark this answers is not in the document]
marinthiercelin: Where does it use it ? I can't find it. I think the tests that anatoly did with drive block encryption included signatures, so I'll ask him.

crypto/keyring_streaming.go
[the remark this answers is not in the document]
wussler: Yeah, i have no naming candidate, and message can indeed be misleading

crypto/keyring_streaming.go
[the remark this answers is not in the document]
wussler: It's buried down into `openpgp.CheckDetachedSignatureAndHash`

crypto/keyring_streaming.go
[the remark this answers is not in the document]
marinthiercelin: Maybe PlainMessageMetadata ?

crypto/sessionkey.go
marinthiercelin: Agreed

crypto/keyring_streaming_test.go
marinthiercelin: I don't know I saw you did this in other tests so I just did the same. Should I remove it ?

crypto/keyring_streaming_test.go
marinthiercelin: True, I'll add that

crypto/sessionkey_streaming_test.go
marinthiercelin: Done

crypto/keyring_streaming_test.go
marinthiercelin: Done

crypto/keyring_streaming_test.go
marinthiercelin: Done

marinthiercelin:
For the changelog, I'll wait that we have reached a final api before documenting the changes.

crypto/keyring_streaming_test.go
wussler: I think it copies when we somehow alter it, if not we can remove it :)

crypto/keyring_message.go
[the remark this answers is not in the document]
marinthiercelin: https://github.com/ProtonMail/gopenpgp/pull/131/commits/d233f833ebed9b96cae9c2aa22c2f0155f313675
That works for you ?

marinthiercelin:
@wussler For the `SessionKey` functions, would you rather have the functions be called `EncryptAndSignStream` and `DecryptAndVerifyStream` to keep consistency with non-streaming functions, or `EncryptStream` and `DecryptStream` to keep consistency with the `KeyRing` stream functions ?

crypto/keyring_message.go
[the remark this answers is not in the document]
wussler: Perfect!

crypto/keyring_streaming.go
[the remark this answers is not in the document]
wussler: Yeah, we can do, expresses the concept well. But then I'd also expect it to be returned from `PlainMessageReader`, maybe with a getter?

crypto/keyring_streaming.go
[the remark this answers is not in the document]
wussler: Let's keep it consistent :+1:

crypto/keyring_streaming.go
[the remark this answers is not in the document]
marinthiercelin: Yes sounds good to me

wussler:
> @wussler For the `SessionKey` functions, would you rather have the functions be called `EncryptAndSignStream` and `DecryptAndVerifyStream` to keep consistency with non-streaming functions, or `EncryptStream` and `DecryptStream` to keep consistency with the `KeyRing` stream functions ?

So, unfortunately I think this was a mistake of mine.
They should have been originally named `Encrypt` and have an optional keyring, but then I didn't consider to encrypt with a session key and sign asymmetrically. So we did not provide that interface.
Now, when required to do so, I "hacked" it in with this `EncryptAndSign` but this breaks the symmetry.
So, my final opinion is to make it `EncryptStream` if this has the "optional" (= not nil) signing key, or if we want to maintain the API as the `SessionKey` then 2 functions `EncryptStream` w/os signing and `EncryptAndSignStream` w/ signing.

marinthiercelin:
The signature/verification keyring is optional so I think `EncryptStream` will fit

crypto/keyring_streaming.go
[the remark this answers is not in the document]
marinthiercelin: Apparently io.Copy copies 32KB at a time, so it's not too bad. It was confirmed that the ios tests included the signature/verification methods, so apparently it's not an issue atm. But it's good to know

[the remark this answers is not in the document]
wussler: Great, thanks!

helper/mobile_stream.go
```diff
@@ -0,0 +1,182 @@
...
+	reader crypto.Reader
+}
+
+// NewGo2IOSReader wraps a native golang Reader to be usable in the ios app runtime (via gomobile).
+func NewGo2IOSReader(reader crypto.Reader) *Go2IOSReader {
+	return &Go2IOSReader{reader}
+}
+
+// Read reads at most <max> bytes from the wrapped Reader and returns the read data as a MobileReadResult.
+func (r *Go2IOSReader) Read(max int) (result *MobileReadResult, err error) {
```
wussler, line 166: Is `MobileReadResult` only used in iOS? If so, wouldn't it better to be `IOSReadResult`?

crypto/keyring_streaming.go
```diff
@@ -0,0 +1,297 @@
...
+	Filename string
+	ModTime  int64
+}
+
+func NewPlainMessageMetadata(isBinary bool, filename string, modTime int64) *PlainMessageMetadata {
+	return &PlainMessageMetadata{IsBinary: isBinary, Filename: filename, ModTime: modTime}
+}
+
+// EncryptStream is used to encrypt data as a Writer.
+// It takes a writer for the encrypted data and returns a writer for the plaintext data
```
wussler, line 38:
```suggestion
// It takes a writer for the encrypted data and returns a WriteCloser for the plaintext data
```

crypto/keyring_streaming.go
```diff
@@ -0,0 +1,297 @@
...
+// It will return an error if the signature is invalid
+// or if the message hasn't been read entirely.
+func (msg *PlainMessageReader) VerifySignature() (err error) {
+	if !msg.readAll {
+		return errors.New("gopenpg: can't verify the signature until the message reader has been read entirely")
+	}
+	if msg.verifyKeyRing != nil {
+		processSignatureExpiration(msg.details, msg.verifyTime)
+		err = verifyDetailsSignature(msg.details, msg.verifyKeyRing)
+	}
```
wussler, line 177: Hmm, I think here we should error if there is no `verifyKeyRing`. Silently "passing" if there is no verifying key can create weird side effects

crypto/sessionkey.go
```diff
@@ -171,65 +172,85 @@ func (sk *SessionKey) EncryptWithCompression(message *PlainMessage) ([]byte, err
...
+	if err != nil {
+		return nil, errors.Wrap(err, "gopenpgp: error in closing encryption writer")
+	}
+	return encBuf.Bytes(), nil
+}
+
+func encryptStreamWithSessionKey(
+	isBinary bool,
+	filename string,
+	modTime uint32,
```
wussler, line 213: Generally we use int64 for the top level API, that internally is cast to uint32 in x/crypto. I'd suggest doing the same here

crypto/sessionkey_streaming.go
```diff
@@ -0,0 +1,105 @@
...
+
+func (w *signAndEncryptWriteCloser) Close() error {
+	if err := w.signWriter.Close(); err != nil {
+		return err
+	}
+	return w.encryptWriter.Close()
+}
+
+// EncryptStream is used to encrypt data as a Writer.
+// It takes a writer for the encrypted data packet and returns a writer for the plaintext data
```
wussler, line 26:
```suggestion
// It takes a writer for the encrypted data packet and returns a writer for the plaintext data.
```

helper/mobile_stream.go
marinthiercelin: No, it's also used in android for the MobileReader interface combined with the Mobile2GoReader.
We have wrappers in both directions.

crypto/keyring_streaming.go
marinthiercelin: Agreed, I added an error

crypto/sessionkey.go
marinthiercelin: That's not the top level api right ? That's an internal method

helper/mobile_stream.go
wussler: OK, then good like this :+1:

crypto/sessionkey.go
wussler: Yeah, true. Since we use it twice and cast it once I'd still prefer to use here int64 :)

marinthiercelin: The problem is that `encryptStream` is also called here: https://github.com/ProtonMail/gopenpgp/blob/feat/streaming-api/crypto/sessionkey.go#L176
and message.Time is already a uint32, so if I put int64 then I have to cast it twice

wussler (approved):

PR #139, opened 2021-07-18
Disregard GNU dummy subkeys when checking (un)locked state and unlocking a crypto.Key

### LINKED ISSUE #138: GNU dummy keys in subkeys cause crypto.Key to be considered both locked and unlocked
I am diagnosing a problem why I cannot log in to the ProtonMail Bridge and Import-Export App, and I was able to reduce the problem to weird behavior in gopenpgp.

When I add this patch to this library, I will get the output, that both `IsLocked()` and `IsUnlocked()` return false. This inconsistency leads to problems in the downstream app.

```patch
diff --git a/crypto/key.go b/crypto/key.go
index 49af11e..5864806 100644
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -135,10 +135,16 @@ func (key *Key) Lock(passphrase []byte) (*Key, error) {
 func (key *Key) Unlock(passphrase []byte) (*Key, error) {
        isLocked, err := key.IsLocked()
        if err != nil {
+               fmt.Printf("(key *Key) Unlock unlocked, err: %+v, %+v\n", isLocked, err)
                return nil, err
        }

        if !isLocked {
+
+               isLocked, err = key.IsLocked()
+               isUnLocked, _ := key.IsUnlocked()
+               fmt.Printf("(key *Key) Unlock isLocked: %t, isUnLocked: %t\n", isUnLocked, isUnLocked)
+
                if passphrase == nil {
                        return key.Copy()
                }

```

output:
```
(key *Key) Unlock isLocked: false, isUnLocked: false
```

I suspect, that (for whatever reason) there might be a subkey, that is considered unlocked(or locked). Due to the nature of this being related to my keys and ProtonMail account, I can not share more details, but I am taking a look if I can identify the issue closer.

In any case, this is with a regular ProtonMail account that had some keys imported into it. So it might be possible there is a corruption introduced there. Regardless, since everything works on the web app and iOS app, I believe there to be an edge-case with regards to this library.

T4cC0re:
Because a GNU dummy key has a `PrivateKey` property, the existing check will
count it as a key to be decrypted. However, a `Decrypt` call on such key will
then yield an error, as it is a dummy key.

This commit addresses this behaviour by summing all valid decrpytable private
keys and returning on their count.
Further, it adapts `Unlock` to ignore any dummy subkey.

Fixes #138


wussler (approved):
Hi @T4cC0re, thanks for your contribution!

PR #145, opened 2021-09-07
Add a synchronous PGP/MIME message decryption function

marinthiercelin:
Add DecryptMIMEMessageSynchronously to decrypt a mime message
and return the decrypted content as a struct rather than
using callbacks.

### DIFF
--- a/crypto/mime.go
+++ b/crypto/mime.go
@@ -10,6 +10,7 @@ import (
 	"github.com/ProtonMail/go-crypto/openpgp"
 	"github.com/ProtonMail/go-crypto/openpgp/packet"
 	gomime "github.com/ProtonMail/go-mime"
+	"github.com/ProtonMail/gopenpgp/v2/constants"
 	"github.com/pkg/errors"
 )
 
@@ -63,8 +64,8 @@ type Attachment struct {
 	Content []byte
 }
 
-// DecryptMIMEMessageSynchronously decrypts a MIME message.
-func (keyRing *KeyRing) DecryptMIMEMessageSynchronously(
+// DecryptMIMEMessageSync decrypts a MIME message.
+func (keyRing *KeyRing) DecryptMIMEMessageSync(
 	message *PGPMessage, verifyKey *KeyRing, verifyTime int64,
 ) (*MIMEMessage, error) {
 	decryptedMessage, err := keyRing.Decrypt(message, verifyKey, verifyTime)
@@ -81,7 +82,8 @@ func (keyRing *KeyRing) DecryptMIMEMessageSynchronously(
 	if err != nil {
 		return nil, err
 	}
-	if sigErr != nil && mimeMessage.SignatureError == nil {
+	if mimeMessage.SignatureError != nil &&
+		mimeMessage.SignatureError.Status == constants.SIGNATURE_NOT_SIGNED {
 		mimeMessage.SignatureError = sigErr
 	}
 	bodyContent, bodyMimeType := body.GetBody()
--- a/crypto/mime.go
+++ b/crypto/mime.go
@@ -52,7 +52,7 @@ func (keyRing *KeyRing) DecryptMIMEMessage(
 }
 
 type MIMEMessage struct {
-	Headers        []string
+	Headers        map[string]string
 	BodyMIMEType   string
 	BodyContent    string
 	Attachments    []*Attachment
@@ -95,7 +95,7 @@ func (keyRing *KeyRing) DecryptMIMEMessageSync(
 			Content: []byte(attachments[i]),
 		})
 	}
-	mimeMessage.Headers = []string{""} // TODO, parse headers
+	mimeMessage.Headers = make(map[string]string) // TODO, parse headers
 	return &mimeMessage, nil
 }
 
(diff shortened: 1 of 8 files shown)


crypto/mime.go
```diff
@@ -46,6 +46,44 @@ func (keyRing *KeyRing) DecryptMIMEMessage(
...
+	Attachments  []*Attachment
+}
+
+type Attachment struct {
+	Header  string
+	Content []byte
+}
+
+// DecryptMIMEMessageSynchronously decrypts a MIME message.
+func (keyRing *KeyRing) DecryptMIMEMessageSynchronously(
```
twiss, line 62: `DecryptMIMEMessageSync`, maybe? (At least that's the convention in Node.js, idk if there's another convention in Kotlin or Android?)

marinthiercelin: I don't think there is such a convention in android, but Sync is shorter while still being clear so it makes sense

crypto/mime.go
```diff
@@ -46,21 +51,69 @@ func (keyRing *KeyRing) DecryptMIMEMessage(
 	callbacks.OnEncryptedHeaders("")
 }
 
+type MIMEMessage struct {
+	Headers        []string
```
marinthiercelin, line 55: @wussler What type would make the most sense for Headers in your view ?
I think we can leave the behavior of not parsing the headers for now, but we should at least get the type right to avoid api changes.

crypto/mime.go
```diff
@@ -33,11 +34,15 @@ func (keyRing *KeyRing) DecryptMIMEMessage(
...
 	}
 
-	body, attachments, attachmentHeaders, err := parseMIME(string(decryptedMessage.GetBinary()), verifyKey)
+	body, attachments, attachmentHeaders, sigErr, err := parseMIME(string(decryptedMessage.GetBinary()), verifyKey)
 	if err != nil {
 		callbacks.OnError(err)
 		return
 	}
+	if sigErr != nil {
+		callbacks.OnError(err) // This is wrong and should use OnVerified instead
```
marinthiercelin, line 43: Should we change this behavior to use OnVerified instead? Or we just consider this function to be legacy and try to deprecate it ?

crypto/mime.go
wussler: Probably a string map?

crypto/mime.go
wussler: I think we should deprecate the callback based one and have only one interface

wussler: Probably if we want to keep the callbacks we should fix `OnVerified`

crypto/signature_collector.go
```diff
@@ -95,17 +97,19 @@ func (sc *SignatureCollector) Accept(
 	}
 	sc.signature = string(buffer)
 	str, _ := ioutil.ReadAll(rawBody)
-	rawBody = bytes.NewReader(str)
+	rawBody = bytes.NewReader([]byte(internal.CanonicalizeAndTrim(string(str))))
```
wussler, line 100: Do we always assume the body to be text?

crypto/mime.go
```diff
@@ -46,21 +51,69 @@ func (keyRing *KeyRing) DecryptMIMEMessage(
...
+	if mimeMessage.SignatureError != nil &&
+		mimeMessage.SignatureError.Status == constants.SIGNATURE_NOT_SIGNED {
+		mimeMessage.SignatureError = sigErr
+	}
+	bodyContent, bodyMimeType := body.GetBody()
+	mimeMessage.BodyContent = bodyContent
+	mimeMessage.BodyMIMEType = bodyMimeType
+	for i := 0; i < len(attachments); i++ {
+		mimeMessage.Attachments = append(mimeMessage.Attachments, &Attachment{
+			Header:  attachmentHeaders[i],
```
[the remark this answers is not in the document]
marinthiercelin: We have one header string per attachment, which contains all the headers and is then parsed by the client.

crypto/signature_collector.go
marinthiercelin: I think so, it seems that binary data is base64 encoded in PGP/MIME. So the whole content is text

crypto/mime.go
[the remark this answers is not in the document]
wussler: Uuuuuh, that seems inconsistent with the other headers property

crypto/mime.go
marinthiercelin: I don't think we need the callbacks, at least in the android client it was anyway blocking until all attachments were processed, I don't know about the iOS client

wussler: IIRC it was iOS adding this originally, so better double check with them

crypto/mime.go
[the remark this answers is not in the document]
marinthiercelin: It's consistent with that though : https://github.com/ProtonMail/gopenpgp/blob/master/crypto/mime.go#L19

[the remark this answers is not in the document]
marinthiercelin: Maybe I should set the body headers to be a single string then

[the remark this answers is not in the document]
wussler: I mean, that would be better from the gomobile perspective. Sucks for the go-based clients though

PR #147, opened 2021-09-23
Fix: use verifyTime in the config time instead of Now()

marinthiercelin:
When decrypting message, we have to use verifyTime in the config
otherwise signatures not valid at verifyTime but valid at Now()
will be seen as valid.

### DIFF
--- a/crypto/keyring_message.go
+++ b/crypto/keyring_message.go
@@ -5,6 +5,7 @@ import (
 	"crypto"
 	"io"
 	"io/ioutil"
+	"time"
 
 	"github.com/ProtonMail/go-crypto/openpgp"
 	"github.com/ProtonMail/go-crypto/openpgp/packet"
@@ -190,6 +191,7 @@ func asymmetricDecrypt(
 		encryptedIO,
 		privateKey,
 		verifyKey,
+		verifyTime,
 	)
 	if err != nil {
 		return nil, err
@@ -215,7 +217,10 @@ func asymmetricDecrypt(
 
 // Core for decryption+verification (all) functions.
 func asymmetricDecryptStream(
-	encryptedIO io.Reader, privateKey *KeyRing, verifyKey *KeyRing,
+	encryptedIO io.Reader,
+	privateKey *KeyRing,
+	verifyKey *KeyRing,
+	verifyTime int64,
 ) (messageDetails *openpgp.MessageDetails, err error) {
 	privKeyEntries := privateKey.entities
 	var additionalEntries openpgp.EntityList
@@ -228,7 +233,19 @@ func asymmetricDecryptStream(
 		privKeyEntries = append(privKeyEntries, additionalEntries...)
 	}
 
-	config := &packet.Config{Time: getTimeGenerator()}
+	config := &packet.Config{
+		Time: func() time.Time {
+			if verifyTime == 0 {
+				/*
+					We default to current time while decrypting and verifying
+					but the caller will remove signature expiration errors later on.
+					See processSignatureExpiration().
+				*/
+				return getNow()
+			}
+			return time.Unix(verifyTime, 0)
+		},
+	}
 
 	messageDetails, err = openpgp.ReadMessage(encryptedIO, privKeyEntries, nil, config)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


crypto/keyring_message.go
```diff
@@ -228,7 +233,14 @@ func asymmetricDecryptStream(
 		privKeyEntries = append(privKeyEntries, additionalEntries...)
 	}
 
-	config := &packet.Config{Time: getTimeGenerator()}
+	config := &packet.Config{
+		Time: func() time.Time {
+			if verifyTime == 0 {
+				return time.Now()
```
marinthiercelin, line 239: I assumed that falling back to device time in this case is ok, since we anyway remove the signature error afterward

twiss: Technically speaking, this is true for `ErrSignatureExpired`, but not `ErrKeyExpired` (not entirely sure if that's intended, but for now it seems safer to fall back to the server time still, if possible)

marinthiercelin: Ah yes I forgot we check that as well.

crypto/keyring_message.go
```diff
@@ -228,7 +233,14 @@ func asymmetricDecryptStream(
 		privKeyEntries = append(privKeyEntries, additionalEntries...)
 	}
 
-	config := &packet.Config{Time: getTimeGenerator()}
+	config := &packet.Config{
+		Time: func() time.Time {
+			if verifyTime == 0 {
+				return getNow()
```
twiss, line 239: This is a bit nitpicky but it may also be worth adding a comment here saying that we ignore signature expiration errors later if `verifyTime == 0`, otherwise readers may get the impression that the intention is that `0` means "current time" (rather than "ignore time")

crypto/keyring_message.go
marinthiercelin: I guess that depends on the intention of verifyTime=0.
Because when verifyTime=<some time> we act as if everything was happening at that time, so key expiration as well.
But when it's 0, we ignore only the expiration of the sig, which is a bit strange.

crypto/keyring_test.go
```diff
@@ -231,3 +233,40 @@ func TestKeyringCapabilities(t *testing.T) {
 	assert.True(t, keyRingTestMultiple.CanVerify())
 	assert.True(t, keyRingTestMultiple.CanEncrypt())
 }
+
+func TestVerificationTime(t *testing.T) {
+	message := NewPlainMessageFromString("Hello")
+	pgp.latestServerTime = 1632312383
```
wussler, line 239: Would you mind resetting the time after the test is done (with a defer) back to `testTime`?

crypto/keyring_message.go
twiss: Yeah, I agree, in openpgp.js, passing `null` would ignore both key expiration and signature expiration. But, that being said, I think ignoring key expiration is a little bit dangerous, so unless someone needs it I wouldn't push very hard for changing this

crypto/keyring_test.go
marinthiercelin: Do I need a defer, if I do `pgp.latestServerTime = testTime` right after ?
Only thing I see that would avoid that is panics.

crypto/keyring_message.go
```diff
@@ -228,7 +233,19 @@ func asymmetricDecryptStream(
 		privKeyEntries = append(privKeyEntries, additionalEntries...)
 	}
 
-	config := &packet.Config{Time: getTimeGenerator()}
+	config := &packet.Config{
+		Time: func() time.Time {
+			if verifyTime == 0 {
+				/*
+					We default to current time while decrypting and verifying
+					but the caller will remove any expiration error later on.
```
twiss, line 241:
```suggestion
					but the caller will remove signature expiration errors later on.
```

crypto/keyring_test.go
wussler: Yeah, and a panic interrupts all tests. It's fine without any defer, as long as there is no fail in the middle.

marinthiercelin: I added a defer, it's more "the golang way"

wussler (approved):

twiss (approved):

PR #149, opened 2021-10-03
feat: return error if verifySignature fail

guilhem:
This prevent hard plumbing and to only use io.Reader interface with
security.

Signed-off-by: Guilhem Lettron <guilhem@barpilot.io>


crypto/keyring_streaming.go
```diff
@@ -174,8 +177,6 @@ func (msg *PlainMessageReader) VerifySignature() (err error) {
 	if msg.verifyKeyRing != nil {
 		processSignatureExpiration(msg.details, msg.verifyTime)
 		err = verifyDetailsSignature(msg.details, msg.verifyKeyRing)
-	} else {
-		err = errors.New("gopenpgp: no verify keyring was provided before decryption")
```
marinthiercelin, line 178: The API is such that if the caller sets the verification key ring to nil, the library should not do any verification. So I don't think we need an error here.

PR #150, opened 2021-10-18
feat: Lock global pgp fields

cquintana92:
This PR adds a `*sync.Mutex` to the `Gopenpgp` type, and uses it to protect concurrent accesses to the fields in the `crypto/time.go` file.

### DIFF
--- a/.github/workflows/ios.yml
+++ b/.github/workflows/ios.yml
@@ -12,10 +12,10 @@ jobs:
     runs-on: macos-latest
 
     steps:
-      - name: Set up xcode 12.2
+      - name: Set up xcode 13.0
         uses: maxim-lobanov/setup-xcode@v1
         with:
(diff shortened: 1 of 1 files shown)


crypto/gopenpgp.go
```diff
@@ -1,14 +1,21 @@
...
 	latestServerTime int64
 	generationOffset int64
+	lock             *sync.Mutex
 }
 
-var pgp = GopenPGP{}
+var pgp = GopenPGP{
+	latestServerTime: 0,
+	generationOffset: 0,
+	lock: &sync.Mutex{},
```
wussler, line 17: I would rather use a `sync.RWMutex` (https://pkg.go.dev/sync#RWMutex) since we have that the memory model is consistent with multiple parallel reads :)

PR #152, opened 2021-11-02
Add Key.IsRevoked command to check key revocation status

wussler:


crypto/key.go
```diff
@@ -276,6 +276,17 @@ func (key *Key) IsExpired() bool {
...
 
+// IsRevoked checks whether the key has a valid revocation signature.
+func (key *Key) IsRevoked() bool {
+	for _, r := range key.entity.Revocations {
+		err := key.entity.PrimaryKey.VerifyRevocationSignature(r)
+		if err == nil {
+			return true
+		}
+	}
+	return false
```
twiss, line 287: go-crypto checks primary key revocation signatures on parsing the key, so this is essentially equivalent to
```suggestion
	return len(key.entity.Revocations) > 0
```
That being said, that's kind of an implementation detail of go-crypto, so probably we should have this function in go-crypto, too, and then just call that here.

Also, this is not the only or most common way to revoke a key, more common is to revoke the primary User ID. So we should also add a function to check if the `Identity` is revoked (and in fact go-crypto doesn't really process User ID revocation signatures at all atm), and then perhaps call both functions here.

wussler: Sounds good to me. We could merge this, have the method in the interface (separate from `isExpired`). Then we fix key revocation in go-crypto and also filter revoked keys from `getEncryptionKeys`, re-implementing here ` isExpired`?

twiss (approved):

PR #154, opened 2021-12-09
Change verification logic for PGP/MIME messages

marinthiercelin:
PGP/MIME can be signed in two places, embedded signature
in the pgp message, and MIME signature in the plaintext.

We now consider the message signature to fail only if both
of these signatures fails.
This means that if at least one of the signatures verify correctly
we return the decrypted message with no errors.


crypto/mime.go
```diff
@@ -48,6 +59,27 @@ func (keyRing *KeyRing) DecryptMIMEMessage(
...
+	}
+	return
+}
+
+func separateSigError(err error) (*SignatureVerificationError, error) {
+	sigErr := &SignatureVerificationError{}
+	isSigError := errors.As(err, sigErr)
+	if isSigError {
+		return sigErr, nil
+	}
```
wussler, line 79:
```suggestion
	if errors.As(err, sigErr) {
		return sigErr, nil
	}
```

crypto/signature_collector.go
```diff
@@ -97,12 +100,15 @@ func (sc *SignatureCollector) Accept(
 	str, _ := ioutil.ReadAll(rawBody)
 	rawBody = bytes.NewReader(str)
 	if sc.keyring != nil {
-		_, err = openpgp.CheckArmoredDetachedSignature(sc.keyring, rawBody, bytes.NewReader(buffer), sc.config)
+		_, err := openpgp.CheckArmoredDetachedSignature(sc.keyring, rawBody, bytes.NewReader(buffer), sc.config)
```
wussler, line 103:
```suggestion
		_, err = openpgp.CheckArmoredDetachedSignature(sc.keyring, rawBody, bytes.NewReader(buffer), sc.config)
```

wussler (approved):
Great PR, wonderful testing. Thanks @marinthiercelin

PR #157, opened 2021-12-15
Add APIs to get the creation time of verified detached signatures

marinthiercelin:


crypto/signature.go
```diff
@@ -159,3 +160,16 @@ func verifySignature(pubKeyEntries openpgp.EntityList, origText io.Reader, signa
...
+
+func (sig *PGPSignature) getCreationTime() (int64, error) {
+	packets := packet.NewReader(bytes.NewReader(sig.Data))
+	p, err := packets.Next()
+	if err != nil {
+		return 0, goerrors.Wrap(err, "gopenpgp: Can't parse signature")
+	}
+	sigPacket, ok := p.(*packet.Signature)
+	if !ok {
+		return 0, errors.New("gopenpgp: non signature packet found")
```
larabr, line 172: what about returning `-1` (or the max int64 value explicitly)? since 0 is a valid timestamp, but -1 would be in the future

twiss: In general, in go it's kind of standard to return zero values when also returning an error, and the caller is supposed to always check the error before using the value (and the value is in that sense meaningless if there is an error), so I think it's fine tbh

marinthiercelin: And in apps it will throw an exception so the value won't be reachable anyway

crypto/keyring_message.go
```diff
@@ -118,6 +118,33 @@ func (keyRing *KeyRing) VerifyDetachedEncrypted(message *PlainMessage, encrypted
...
+		return 0, err
+	}
+	return signature.getCreationTime()
+}
+
+// GetVerifiedEncryptedSignatureTimestamp verifies a PlainMessage
+// with a PGPMessage containing an encrypted detached signature
+// returns the creation time of the signature if it succeeds
+// and returns a SignatureVerificationError if fails.
+func (keyRing *KeyRing) GetVerifiedEncryptedSignatureTimestamp(message *PlainMessage, encryptedSignature *PGPMessage, decryptionKeyRing *KeyRing, verifyTime int64) (int64, error) {
```
larabr, line 136: Do we have a use-case in mind for this? It's not needed by my side

marinthiercelin: I don't think so, it's more for consistency but we can ditch it

larabr (approved):

crypto/keyring_message.go
```diff
@@ -118,6 +118,17 @@ func (keyRing *KeyRing) VerifyDetachedEncrypted(message *PlainMessage, encrypted
 	return keyRing.VerifyDetached(message, signature, verifyTime)
 }
 
+// GetVerifiedSignatureTimestamp verifies a PlainMessage with a detached PGPSignature
+// returns the creation time of the signature if it succeeds
+// and returns a SignatureVerificationError if fails.
+func (keyRing *KeyRing) GetVerifiedSignatureTimestamp(message *PlainMessage, signature *PGPSignature, verifyTime int64) (int64, error) {
+	err := keyRing.VerifyDetached(message, signature, verifyTime)
```
wussler, line 125: Here it verifies all the signatures, but then later we return the timestamp of the first one if any of them succeeds. We should return the timestamp of the succeeding one

wussler (changes requested):

crypto/keyring_message.go
marinthiercelin: Can you have several sigs in a detached signature ?
If yes, how would I find out which one succeeded ?

wussler: Yeah, it can happen. A simple solution here is to accept just one, since the crypto API `CheckDetachedSignatureAndHash` returns only the signer entity, and not the signature packet.

So basically instead of using `signature.getCreationTime()` you can just check if the first packet is a signature, then serialise just that one and verify it with `verifySignature`

marinthiercelin: Wouldn't this lead to inconsistency between VerifyDetached and GetVerifiedSignatureTimestamp ?
Some sigs would be accepted by the first but not the other one

wussler: Indeed. The other solution I see is to do the same but manually iterating over the signature packets, verifying one by one and returning the verifying one's timestamp

marinthiercelin: Ok I tried going over each packet in a loop and return the time stamp for the first packet that verifies

CHANGELOG.md
```diff
@@ -4,6 +4,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## Unreleased
+
+### Added
+- Function to verify a detached signature and get it's creation time:
```
wussler, line 10: \
```suggestion
- Function to verify a detached signature and get its creation time: 
```

wussler (approved):

PR #170, opened 2022-02-28
Create `SplitMessage()` to replace `SeparateKeyAndData(...int)`

twiss:
Keep `SeparateKeyAndData(_ int, _ int)` for backwards compatibility with go-mobile bindings.
Deprecate `SeparateKeyAndData` in favor of `SplitMessage`.

### DIFF
--- a/crypto/message.go
+++ b/crypto/message.go
@@ -140,7 +140,7 @@ func NewPGPSplitMessageFromArmored(encrypted string) (*PGPSplitMessage, error) {
 		return nil, err
 	}
 
-	return message.SeparateKeyAndData()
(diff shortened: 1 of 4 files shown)


crypto/message.go
```diff
@@ -352,6 +352,13 @@ Loop:
 	}, nil
 }
 
+// SeparateKeyAndData splits the message into key and data packet(s).
+// Parameters are for backwards compatibility and are unused.
+// Deprecated in favor of SplitMessage().
```
wussler, line 357:
```suggestion
// Deprecated: use SplitMessage().
```

wussler (approved):

PR #178, opened 2022-04-18
Allow key generation with only name or only email set.

### LINKED ISSUE #177: Allow empty email for key generation
Dear gopenpgp maintainers,

thank you very much for your library as it makes OpenPGP easily usable in Go. I am using it to implement "XEP-0373: OpenPGP for XMPP" but I have an issue: The XEP requires the User ID to be in the format "xmpp:user@example.org" but gopenpgp/crypto requires an email to be set so I can only create "xmpp:user@example.org \<xmpp:user@example.org\>". Could you add a function to create keys more flexible to make the library also useful outside of the email use case? Or just change https://github.com/ProtonMail/gopenpgp/blob/e04eb20cba8a624dc2f7a8973046e898ad27cb26/crypto/key.go#L441 to accept empty email or name values.

mdosch:
This allows generating keys for other use cases
than email, e.g. for Ox "OpenPGP for XMPP"
XEP-0373.

Fixes #177

### DIFF
--- a/crypto/key.go
+++ b/crypto/key.go
@@ -81,6 +81,15 @@ func GenerateRSAKeyWithPrimes(
 func GenerateKey(name, email string, keyType string, bits int) (*Key, error) {
 	return generateKey(name, email, keyType, bits, nil, nil, nil, nil)
 }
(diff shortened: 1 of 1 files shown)


wussler:
Merged with #180

PR #199, opened 2022-11-16
Detached sign text messages with signature type text

### LINKED ISSUE #198: gopenpgp makes binary detached signatures even when signing text
@twiss wrote in   https://github.com/ProtonMail/gosop/issues/10#issuecomment-1315751090 :

> gopenpgp […] creates a binary signature even when signing text

twiss:
Fix #198. This is required to fix https://github.com/ProtonMail/gosop/issues/10. Also, `SignDetached()` is used for clearsigned messages as well, where using binary type signatures is a bit strange. This fixes that as well.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Changed
 - Updated `github.com/ProtonMail/go-mime` to latest versions, which cleans up uneeded dependencies. And fix an issue with PGP/MIME messages with non standard encodings.
 - Sanitize strings returned in `MIMECallbacks.OnBody()` and `PlainMessage.GetString()`. Strings that have non utf8 characters will be sanitized to have the "character unknown" character : � instead.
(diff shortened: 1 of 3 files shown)


marinthiercelin:
Should we add a text version to the streaming api as well ? 
KeyRing.SignDetachedStreaming will only produce binary signatures afaik.

twiss:
Hmm. Yeah, it's a good point, but I think the streaming functions are most commonly used for large files, i.e. typically binary, so perhaps it's not as critical?

marinthiercelin:
> Hmm. Yeah, it's a good point, but I think the streaming functions are most commonly used for large files, i.e. typically binary, so perhaps it's not as critical?

Right, at least in pm we don't have a usecase for it, so we can add it later if someone requests it.

marinthiercelin:
I added some type checks for the unit tests : https://github.com/ProtonMail/gopenpgp/pull/201/files

marinthiercelin (commented):
Can you update the changelog please ?

twiss:
> Doesn't that mean that VerifyDetached should look at the signature type and apply the canonicalization based on that ?

go-crypto does so :)

marinthiercelin (approved):

PR #200, opened 2022-11-16
Don't trim trailing spaces from non-clearsigned text messages

twiss:


marinthiercelin:
For exisiting detached signatures that were made from NewPlainMessageString(), won't that break verification ?
If the signature was made with spaces trimmed, but now verification is done without ?

twiss:
Yeah, for existing signatures, the application will have to trim the spaces, so it's a breaking change, we'll have to communicate that.

marinthiercelin (commented):
Some additional tests: https://github.com/ProtonMail/gopenpgp/pull/202
Can you also update the changelog please ?

marinthiercelin (approved):

PR #207, opened 2023-01-09
Clean dependencies

wussler:


marinthiercelin:
Can you run go mod tidy ? I don't think the go-mobile dependency is needed for gopenpgp, and go mod tidy should clean it up

wussler:
That makes the pipeline for iOS and Android fail

marinthiercelin (approved):

wussler:
The bothering part was that it was in the go.sum but not in the go.mod

Issue #210, opened 2023-01-25
Upgrade from 2.4.10 to 2.5.1 breaks my project: not enough arguments in call to packet.SerializeSymmetricallyEncrypted


meshenka:
```
# github.com/ProtonMail/gopenpgp/v2/crypto
../../go/pkg/mod/github.com/!proton!mail/gopenpgp/v2@v2.5.1/crypto/sessionkey.go:242:105: not enough arguments in call to packet.SerializeSymmetricallyEncrypted
	have (io.Writer, packet.CipherFunction, []byte, *packet.Config)
	want (io.Writer, packet.CipherFunction, bool, packet.CipherSuite, []byte, *packet.Config)
```

marinthiercelin:
I think you might be using a version of github.com/ProtonMail/go-crypto which is more recent than the one used in gopenpgp

meshenka:
nice i will give it a try!

meshenka:
this issue come with 2.5.10

mdosch:
Same for me (go-sendxmpp):

```
go get -u .
go: downloading github.com/ProtonMail/gopenpgp/v2 v2.5.1
go: downloading github.com/ProtonMail/go-crypto v0.0.0-20230124153114-0acdc8ae009b
go: downloading github.com/stretchr/testify v1.7.0
go: upgraded github.com/ProtonMail/go-crypto v0.0.0-20230117203413-a47887b8f098 => v0.0.0-20230124153114-0acdc8ae009b
go: upgraded github.com/ProtonMail/gopenpgp/v2 v2.5.0 => v2.5.1
```

```
go build .
# github.com/ProtonMail/gopenpgp/v2/crypto
../go/pkg/mod/github.com/!proton!mail/gopenpgp/v2@v2.5.1/crypto/sessionkey.go:242:105: not enough arguments in call to packet.SerializeSymmetricallyEncrypted
	have (io.Writer, packet.CipherFunction, []byte, *packet.Config)
	want (io.Writer, packet.CipherFunction, bool, packet.CipherSuite, []byte, *packet.Config)
```

mdosch:
Thanks, 2.5.2 works with latest go-crypto. :)

PR #211, opened 2023-01-25
Upgrade go crypto and adapt SEIPD

wussler:


wussler:
This issue closes #210

meshenka (approved):

crypto/sessionkey.go
```diff
@@ -239,7 +239,15 @@ func encryptStreamWithSessionKey(
 	signEntity *openpgp.Entity,
 	config *packet.Config,
 ) (encryptWriter, signWriter io.WriteCloser, err error) {
-	encryptWriter, err = packet.SerializeSymmetricallyEncrypted(dataPacketWriter, config.Cipher(), sk.Key, config)
+	encryptWriter, err = packet.SerializeSymmetricallyEncrypted(
+		dataPacketWriter,
+		config.Cipher(),
```
marinthiercelin, line 244: Why do we need the cipher both in here and in the ciphersuite ?

wussler: In this case they're taken from the config, but in a key you could have different preferences for AEAD and for the normal cipher. So we need a generic interface that allows to pass different settings here

See [the SEIPD v1 preferences](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-07.html#name-preferred-symmetric-ciphers) vs [the SEIPD v2 preferences](https://www.ietf.org/archive/id/draft-ietf-openpgp-crypto-refresh-07.html#name-preferred-aead-ciphersuites)

marinthiercelin: But for a single call to SerializeSymetricallyEncrypted, you would use only one cipher right ?

wussler: Yes, correct, the picked cipher depends whether AEAD is enabled or not

marinthiercelin (approved):

PR #214, opened 2023-02-17
Update go crypto, and improve signature verification

twiss:
- Update go-crypto
- More strictly verify detached signatures: reject detached signatures from revoked and expired keys.
- In `GetVerifiedSignatureTimestamp`, use the new `VerifyDetachedSignatureAndHash` to get the verified signature, instead of parsing the signature packets manually to get the timestamp.

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -119,7 +119,7 @@ func verifyDetailsSignature(md *openpgp.MessageDetails, verifierKey *KeyRing) er
 }
 
 // verifySignature verifies if a signature is valid with the entity list.
-func verifySignature(pubKeyEntries openpgp.EntityList, origText io.Reader, signature []byte, verifyTime int64) error {
+func verifySignature(pubKeyEntries openpgp.EntityList, origText io.Reader, signature []byte, verifyTime int64) (*packet.Signature, error) {
 	config := &packet.Config{}
 	if verifyTime == 0 {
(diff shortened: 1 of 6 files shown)


crypto/signature.go
```diff
@@ -144,18 +147,15 @@ func verifySignature(pubKeyEntries openpgp.EntityList, origText io.Reader, signa
...
-		signer, err = openpgp.CheckDetachedSignatureAndHash(pubKeyEntries, origText, signatureReader, allowedHashes, config)
-		if err != nil {
-			return newSignatureFailed()
-		}
+		sig, signer, err = openpgp.VerifyDetachedSignatureAndHash(pubKeyEntries, origText, signatureReader, allowedHashes, config)
 	}
 
-	if signer == nil {
-		return newSignatureFailed()
+	if err != nil || sig == nil || signer == nil {
```
marinthiercelin, line 156: Here we are swallowing the error (we were already doing it though), shouldn't we wrap it into the returned error ?

marinthiercelin (commented):
Can you update the changelog as well please ?

crypto/signature.go
twiss: Yeah, I'd tend to agree, but I'd do it in a separate PR, since it's not strictly speaking related to this go-crypto update

marinthiercelin (approved):

PR #220, opened 2023-03-07
Wrap the cause of signature verification errors

marinthiercelin:
Instead of swallowing verification errors, we use errors.Join to keep all the errors.


twiss:
`errors.Join` is Go 1.20+ only, do we already use that everywhere? If not, or perhaps regardless, it might be cleaner to give `newSignatureFailed()` an `err` param, which is then stored in the `SignatureVerificationError` object, and added to the message in its `Error()` method, and you could then also implement `Unwrap()` to return it, so that `errors.Is` can be used to check the type of signature failure: https://pkg.go.dev/errors.

marinthiercelin:
> `errors.Join` is Go 1.20+ only, do we already use that everywhere? If not, or perhaps regardless, it might be cleaner to give `newSignatureFailed()` an `err` param, which is then stored in the `SignatureVerificationError` object, and added to the message in its `Error()` method, and you could then also implement `Unwrap()` to return it, so that `errors.Is` can be used to check the type of signature failure: https://pkg.go.dev/errors.

Oh good catch, no we're not using go 1.20 everywhere yet unfortunately.
I'll try to implement your proposal.

CHANGELOG.md
```diff
@@ -19,6 +19,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 - Update `github.com/ProtonMail/go-crypto` to the latest version
 - More strictly verify detached signatures: reject detached signatures from revoked and expired keys.
 - In `GetVerifiedSignatureTimestamp`, use the new `VerifyDetachedSignatureAndHash` function to get the verified signature, instead of parsing the signature packets manually to get the timestamp.
+- The `SignatureVerificationError` struct now has a `Cause error` parameter, which is returned by the the Unwrap function.
```
twiss, line 22:
```suggestion
- The `SignatureVerificationError` struct now has a `Cause error` field, which is returned by the the Unwrap function.
```

twiss (approved):
Tiny nitpick below, but lgtm :+1:

PR #224, opened 2023-03-15
Wrap the cause of signature verification errors.

marinthiercelin:
Instead of swallowing the cause of verification errors, we use error wrapping to communicate the cause to the caller.


marinthiercelin:
Last PR was merged in the wrong branch, so it didn't make it to main.

twiss (approved):

Issue #231, opened 2023-04-07
crypto.verifySignature is broken for signatures with expiration


DmitriyMV:
`verifySignature` is broken if it tries to retry with the actual verification time. It doesn't reset `origText` reader, so it has nothing to read on second attempt.

Relevant [parts](https://github.com/ProtonMail/gopenpgp/blob/main/crypto/signature.go#L245-L262).

marinthiercelin:
Hey, thanks for opening the issue.
That seems indeed to be broken.
I'm a bit confused why you would revert your project to v2.5.2 though, the issue seems older than this version.
Do you have any example data that I could use to trigger the bug ?

DmitriyMV:
> I'm a bit confused why you would revert your project to v2.5.2 though, the issue seems older than this version.

`verifySignature` is indeed broken in 2.5.2 (it doesn't return an error if signature expired) but we are not hitting it, because we have check called [IsExpired](https://github.com/siderolabs/go-api-signature/blob/69886dcc1343561add3b4b86ef160e0a1876d97f/pkg/pgp/key.go#L98-L119) which is doing what it must. So that's a reason for revert. 

> Do you have any example data that I could use to trigger the bug ?

[Sure, here it is.](https://go.dev/play/p/yaPW-ShcZUr)

marinthiercelin:
Hey, It's indeed tricky to fix while keeping the API.

I think, as a short term fix, we should go with a variant of your second approach.
If `origText` supports `Seek` use it and verify again.
But if it does not support Seek, I don't think we should read it all in a buffer, that could be quite bad for memory usage.
Instead I think we should just not reattempt the verification.
That means that if you verify with `VerifyDetachedStream` with a non seekable stream, the verification might fail.
So it's only a partial fix.

For a complete fix, I think we should instead rethink how we do the verification, and manually check the dates for expiration, rather than reattempting verification with a different time.

If you want to open a PR for either the partial fix, or look into the complete fix, it would be greatly appreciated.

DmitriyMV:
Opened an PR for partial fix.

marinthiercelin:
Thanks a lot for your submission @DmitriyMV

smira:
@DmitriyMV is there something else which needs to fixed here? as it seems that the verification is still broken?

DmitriyMV:
@smira It's a partial fix in a sense that it only supports messages which have `Seek` method. It may still trigger for non Seek-able streams, but for the most part it is good enough.

PR #233, opened 2023-04-12
Reset origText in verifySignature before the retry

DmitriyMV:
This is partial fix for #231

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -251,7 +251,7 @@ func verifySignature(
 
 	sig, signer, err := openpgp.VerifyDetachedSignatureAndHash(pubKeyEntries, origText, signatureReader, allowedHashes, config)
 
-	if sig != nil && signer != nil && (errors.Is(err, pgpErrors.ErrSignatureExpired) || errors.Is(err, pgpErrors.ErrKeyExpired)) {
+	if sig != nil && signer != nil && (errors.Is(err, pgpErrors.ErrSignatureExpired) || errors.Is(err, pgpErrors.ErrKeyExpired)) { //nolint:nestif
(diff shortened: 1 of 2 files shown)


crypto/signature.go
```diff
@@ -261,12 +261,22 @@ func verifySignature(
 				return time.Unix(verifyTime, 0)
 			}
 
+			seeker, ok := origText.(io.ReadSeeker)
+			if !ok {
+				return nil, newSignatureFailed(errors.New("message reader do not support seeking, cannot retry signature verification"))
```
marinthiercelin, line 266:
```suggestion
				return nil, newSignatureFailed(errors.Wrap(err, "gopenpgp: message reader do not support seeking, cannot retry signature verification"))
```
We should prefix the error message with the module name. And I would include the original error.

crypto/signature.go
```diff
@@ -261,12 +261,22 @@ func verifySignature(
...
+			}
+
+			_, err = seeker.Seek(0, io.SeekStart)
+			if err != nil {
+				return nil, newSignatureFailed(err)
```
marinthiercelin, line 271: We should wrap errors with the context.
```suggestion
				err = errors.Wrap(err, "gopenpgp: could not rewind the data reader.")
				return nil, newSignatureFailed(err)
```
Can you also add this to the error from `signatureReader.Seek(0, io.SeekStart)` below ? It seems we forgot to wrap it.

DmitriyMV: Thanks fixed!

crypto/signature.go
DmitriyMV: Fixed!

marinthiercelin (approved):

PR #240, opened 2023-05-12
packet.Config.Time should return time with 999999999 nsec

DmitriyMV:
Currently verifySignature accepts seconds for verification, but internally it uses time.Time for comparison. That means if we created openpgp.Entity with default time field, test can fail with "key expired" error because PublicKey.KeyExpired will think that key creation happened before verifyTime when in reality we just lost the nsec precision.

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -240,7 +240,7 @@ func verifySignature(
 		}
 	} else {
 		config.Time = func() time.Time {
-			return time.Unix(verifyTime+internal.CreationTimeOffset, 0)
+			return time.Unix(verifyTime+internal.CreationTimeOffset, 999999999)
 		}
 	}
 
@@ -258,7 +258,7 @@ func verifySignature(
 			// Maybe the creation time offset pushed it over the edge
 			// Retry with the actual verification time
 			config.Time = func() time.Time {
-				return time.Unix(verifyTime, 0)
+				return time.Unix(verifyTime, 999999999)
 			}
 
 			seeker, ok := origText.(io.ReadSeeker)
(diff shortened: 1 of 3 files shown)


DmitriyMV:
Relevant test added that covers both this and #231

twiss:
Hello :wave: Timestamps in OpenPGP only have second precision. I assume that if you'd serialize and parse the key before using it, this issue would go away - which implies that the fix should be something else, i.e. round down the current time when generating keys (and signatures). Though, that would probably need to be done in go-crypto, not gopenpgp.

twiss:
The thing is that both [`pk.CreationTime`](https://github.com/ProtonMail/go-crypto/blob/58e86b2947569471f4657525aeb6e835e7c7d3eb/openpgp/packet/public_key.go#L196) and [`sig.CreationTime`](https://github.com/ProtonMail/go-crypto/blob/58e86b2947569471f4657525aeb6e835e7c7d3eb/openpgp/packet/signature.go#L324), when parsed, come from an integer-precision timestamp (see [section 3.5 of RFC 4880](https://www.rfc-editor.org/rfc/rfc4880#section-3.5)). So in that case, truncating them should be unnecessary - but when the objects are created directly (e.g. in [`NewRSAPublicKey`](https://github.com/ProtonMail/go-crypto/blob/58e86b2947569471f4657525aeb6e835e7c7d3eb/openpgp/packet/public_key.go#L76), etc), rather than parsed, we should probably truncate the times, for consistency, yeah. Or, alternatively, we could do it in [`config.Now`](https://github.com/ProtonMail/go-crypto/blob/58e86b2947569471f4657525aeb6e835e7c7d3eb/openpgp/packet/config.go#L138) - that might be simplest.

DmitriyMV:
[openpgp.Entity.EncryptionKey](https://github.com/ProtonMail/go-crypto/blob/58e86b2947569471f4657525aeb6e835e7c7d3eb/openpgp/keys.go#L121) is public method which accepts non-truncated time. I suppose there are also other ways to pass nsec time to `KeyExpired` functions. Truncating there should safe-guard against that getting in.

DmitriyMV:
@twiss https://github.com/ProtonMail/go-crypto/pull/168 like this?

Issue #213, opened 2023-01-28
No helper.EncryptSignBinaryMessageArmored


wneessen:
I am currently working on a Middleware for encrypting/signing mails with go-mail and was wondering why the helper packages does not provide a `EncryptSignBinaryMessageArmored` method. Is that intentionally or has it just never been needed? If so, it would be great to have that.

twiss:
Hey :wave: The helper functions have indeed mostly been added on an as-needed basis, and are intended to cover the most common use-cases, while the lower-level API can cover more uncommon use cases. That being said, we are planning to re-work the API to make the high-level API more composable as well, and support more combinations of operations.

PR #246, opened 2023-06-21
Decrypt verify message armored

snowstorm399:
Hi,

Not sure if this is worth the effort anymore, but it's not quite straight through to do a decrypt and verify of binary data right now. This PR adds that functionality.

I read [here](https://github.com/ProtonMail/gopenpgp/issues/213) that an overhaul is going on so could not be worth the effort. If it is, I'll make another PR to add EncryptSignBinaryMessageArmored


snowstorm399:
Need to fix the tests.

PR #249, opened 2023-08-24
New function to quick check if a session key can decrypt a encrypted data packet

lubux:
This pull request adds a new function `QuickCheckDecrypt` to the `helper` package.
The function allows to check with high probability if a session key can decrypt a data packet given its 24 byte prefix.
It only works for `SEIPDv1` data packets that use `AES` as a cipher.

### DIFF
--- a/helper/decrypt_check.go
+++ b/helper/decrypt_check.go
@@ -17,6 +17,8 @@ func supported(cipher packet.CipherFunction) bool {
 	switch cipher {
 	case packet.CipherAES128, packet.CipherAES192, packet.CipherAES256:
 		return true
+	case packet.CipherCAST5, packet.Cipher3DES:
+		return false
 	}
 	return false
 }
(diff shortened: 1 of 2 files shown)


helper/decrypt_check.go
```diff
@@ -0,0 +1,101 @@
...
+		encryptedData[blockSize+ind] ^= blockBuffer[ind]
+	}
+	for ind := range blockBuffer {
+		blockBuffer[ind] = 0
+	}
+	// Decrypt the first block
+	blockCipher.Encrypt(blockBuffer, blockBuffer)
+	for ind := range blockBuffer {
+		encryptedData[ind] ^= blockBuffer[ind]
+	}
```
twiss, line 89: Just to note: it might be simpler to create a CFB instance, which would do this for you? But, not critical, we can merge it as-is and change it later

twiss (approved):

helper/decrypt_check.go
lubux: Good point, I agree. It is slightly more expensive (1 more buffer and 2 copies), but should not be noticeable and there is no code duplication. 

[f1e673a](https://github.com/ProtonMail/gopenpgp/pull/249/commits/f1e673ab63f95fa3c181c9b1b419e76d40b1a164) integrates the change.

PR #275, opened 2024-04-04
Add parsing and serializing interfaces for keyrings

wussler:
This PR adds an API to serialize KeyRings to binary data:
```go
func (keyRing *KeyRing) Serialize() ([]byte, error)
```
and an API to parse KeyRings from binary data:
```go
func NewKeyRingFromBinary(binKeys []byte) (*KeyRing, error)
```

### DIFF
--- a/crypto/keyring.go
+++ b/crypto/keyring.go
@@ -49,6 +49,30 @@ func (keyRing *KeyRing) AddKey(key *Key) error {
 	return nil
 }
 
+// NewKeyRingFromBinary creates a new keyring with all the keys contained in the unarmored binary data.
+// Note that it accepts only unlocked or public keys, as KeyRing cannot contain locked keys.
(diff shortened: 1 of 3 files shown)


crypto/keyring.go
```diff
@@ -49,6 +49,30 @@ func (keyRing *KeyRing) AddKey(key *Key) error {
...
+	}
+
+	keyring := &KeyRing{}
+	for _, entity := range entities {
+		key, err := NewKeyFromEntity(entity)
+		if err != nil {
+			return nil, errors.Wrap(err, "gopenpgp: error in reading keyring")
+		}
+
+		err = keyring.AddKey(key)
```
lubux, line 67: Nitpick could be:

```
if err = keyring.AddKey(key); err != nil {
...
}
```

lubux (approved):
LGTM

twiss (approved):

PR #284, opened 2024-06-11
v3: Rework Armor checksum handling

lubux:
GopenPGP v3 did not produce any armor checksum as recommended by the crypto refresh. Unfortunately, a popular OpenPGP library fails to parse armored messages without a checksum in certain scenarios.

This MR adds armor checksums back per default, but tries to avoid them when generating crypto refresh messages. i.e., generated by v6 keys.

In GopenPGP v3, armor checksums were not produced, following recommendations from the OpenPGP crypto refresh. Unfortunately, a widely used OpenPGP library fails to parse armored messages that lack a checksum in certain scenarios although they should be optional according to the official RFC.

This merge request (MR) reinstates armor checksums by default to ensure compatibility. However, it tries to omit the checksums when generating messages for the crypto-refresh, i.e., those generated with v6 keys.

Changes:

- API to armor data with the option to remove the checksum
- All armor functions append a checksum per default for compatibility with certain libraries although the crypto-refresh advises not to.
- Encryption and Sign handle now append a checksum when armoring. If the produced OpenPGP packets are crypto-refresh packets, the checksum is not appended as mandated by the crypto-refresh.

### DIFF
--- a/constants/armor.go
+++ b/constants/armor.go
@@ -3,6 +3,7 @@ package constants
 
 // Constants for armored data.
 const (
+	DoChecksum         = true
 	ArmorHeaderEnabled = false // can be enabled for debugging at compile time only
 	ArmorHeaderVersion = "GopenPGP " + Version
 	ArmorHeaderComment = "https://gopenpgp.org"
--- a/constants/armor.go
+++ b/constants/armor.go
@@ -3,12 +3,18 @@ package constants
 
 // Constants for armored data.
 const (
-	DoChecksum         = true
-	ArmorHeaderEnabled = false // can be enabled for debugging at compile time only
-	ArmorHeaderVersion = "GopenPGP " + Version
-	ArmorHeaderComment = "https://gopenpgp.org"
-	PGPMessageHeader   = "PGP MESSAGE"
-	PGPSignatureHeader = "PGP SIGNATURE"
-	PublicKeyHeader    = "PGP PUBLIC KEY BLOCK"
-	PrivateKeyHeader   = "PGP PRIVATE KEY BLOCK"
+	// ArmorChecksumSetting defines the default behavior for adding an armor checksum
+	// to an armored message.
+	//
+	// If set to true, an armor checksum is added to the message.
+	//
+	// If set to false, no armor checksum is added.
+	ArmorChecksumSetting = true
+	ArmorHeaderEnabled   = false // can be enabled for debugging at compile time only
+	ArmorHeaderVersion   = "GopenPGP " + Version
+	ArmorHeaderComment   = "https://gopenpgp.org"
+	PGPMessageHeader     = "PGP MESSAGE"
+	PGPSignatureHeader   = "PGP SIGNATURE"
+	PublicKeyHeader      = "PGP PUBLIC KEY BLOCK"
+	PrivateKeyHeader     = "PGP PRIVATE KEY BLOCK"
 )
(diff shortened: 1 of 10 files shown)


crypto/key.go
```diff
@@ -433,6 +433,13 @@ func (key *Key) ToPublic() (publicKey *Key, err error) {
 	return
 }
 
+func (key *Key) isVersionSix() bool {
```
twiss, line 436: Normally I'm in favor of writing things out fully, but for this we normally refer to them as "V6" so I would call this `isV6`, also for consistency with go-crypto's `UpgradeToV6` etc.

crypto/message.go
```diff
@@ -36,6 +36,8 @@ type PGPMessage struct {
 	DetachedSignature []byte
 	// detachedSignatureIsPlain indicates if the detached signature is not encrypted.
 	detachedSignatureIsPlain bool
+	// Signals that no armor checksum must be appended when armoring
+	forceNoArmorChecksum bool
```
twiss, line 40: How about `omitArmorChecksum`?

crypto/sign_handle.go
```diff
@@ -133,6 +131,18 @@ func (sh *signatureHandle) validate() error {
 	return nil
 }
 
+func (sh *signatureHandle) doArmorChecksum() bool {
```
twiss, line 134: (Same as above here)

crypto/encryption_handle.go
```diff
@@ -150,6 +151,39 @@ func (eh *encryptionHandle) validate() error {
 	return nil
 }
 
+// doArmorChecksum determines if an armor checksum should be appended or not.
+// The OpenPGP Crypto-Refresh mandates that no checksum should be appended with the new packets.
+func (eh *encryptionHandle) doArmorChecksum() bool {
+	encryptionConfig := eh.profile.EncryptionConfig()
+	if encryptionConfig.AEADConfig == nil {
+		return constants.DoChecksum
```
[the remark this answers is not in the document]
lubux: Yeah the variable name is not well chosen here. The idea is that we have a global setting to disable the checksum in the future. 
I renamed the variable and do the check in the beginning of the function.
```go
if !constants.ArmorChecksumSetting {
  // If the default behavior is no checksum, we can ignore
  // the logic for the crypto-refresh check.
  return false
}
```

lubux:
Thanks for the review. I addressed the comments.

constants/armor.go
```diff
@@ -3,6 +3,7 @@ package constants
 
 // Constants for armored data.
 const (
+	DoChecksum         = true
```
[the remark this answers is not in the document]
lubux: See comment below.

constants/armor.go
```diff
@@ -3,12 +3,18 @@ package constants
...
-	PGPSignatureHeader = "PGP SIGNATURE"
-	PublicKeyHeader    = "PGP PUBLIC KEY BLOCK"
-	PrivateKeyHeader   = "PGP PRIVATE KEY BLOCK"
+	// ArmorChecksumSetting defines the default behavior for adding an armor checksum
+	// to an armored message.
+	//
+	// If set to true, an armor checksum is added to the message.
+	//
+	// If set to false, no armor checksum is added.
+	ArmorChecksumSetting = true
```
twiss, line 12: OK, I understand a bit better now what the intention is :D How about `ArmorChecksumEnabled`, to match the next one?

Other than that, LGTM :+1:

twiss (approved):

PR #295, opened 2024-09-06
docs: Remove redundant example code

pascal-hofmann:
The code also generates curve25519 v4 keys at the moment.

### DIFF
--- a/README.md
+++ b/README.md
@@ -282,8 +282,6 @@ rsaKeyHigh, err := keyGenHandle.GenerateKeyWithSecurity(constants.HighSecurity)
 keyGenHandle = pgpDefault.KeyGeneration().AddUserId(name, email).New()
 // Generates curve25519 v4 keys.
 ecKey, err := keyGenHandle.GenerateKey()
-// Generates curve448 v4 keys.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -282,7 +282,7 @@ rsaKeyHigh, err := keyGenHandle.GenerateKeyWithSecurity(constants.HighSecurity)
 keyGenHandle = pgpDefault.KeyGeneration().AddUserId(name, email).New()
 // Generates curve25519 v4 keys.
 ecKey, err := keyGenHandle.GenerateKey()
-// Generates curve448 v4 keys.
+// Also generates curve25519 v4 keys.
 ecKeyHigh, err := keyGenHandle.GenerateKeyWithSecurity(constants.HighSecurity)
```
twiss, line 286:
```suggestion
```

twiss (commented):
Hey :wave: Good catch! However, probably it would make more sense to just remove this example for now (since it's pretty useless as-is), until we can make it true and generate v4 Curve448 keys. RFC9580 (which introduces Curve448) is out now, but we might still want to wait a bit for support to become more widespread.

README.md
pascal-hofmann: I removed the example.

twiss (approved):
:+1: Thanks!

PR #299, opened 2024-09-20
feat: Add SHA256 fingerprint getter

lubux:
Add a getter for the primary key SHA256 fingerprint.


crypto/key_test.go
```diff
@@ -374,6 +374,7 @@ func TestGetSHA256FingerprintsV4(t *testing.T) {
 	assert.Len(t, sha256Fingerprints, 2)
 	assert.Exactly(t, "d9ac0b857da6d2c8be985b251a9e3db31e7a1d2d832d1f07ebe838a9edce9c24", sha256Fingerprints[0])
 	assert.Exactly(t, "203dfba1f8442c17e59214d9cd11985bfc5cc8721bb4a71740dd5507e58a1a0d", sha256Fingerprints[1])
+	assert.Exactly(t, "d9ac0b857da6d2c8be985b251a9e3db31e7a1d2d832d1f07ebe838a9edce9c24", publicKey.GetSHA256Fingerprint())
```
twiss, line 377: This is of course extremely nitpicky but if we ever want to cherry-pick changes between the two branches it might help if the patch is exactly the same as in https://github.com/ProtonMail/gopenpgp/pull/296:
```suggestion

	assert.Exactly(t, "d9ac0b857da6d2c8be985b251a9e3db31e7a1d2d832d1f07ebe838a9edce9c24", publicKey.GetSHA256Fingerprint())
```

twiss (approved):

PR #303, opened 2024-10-29
Remove signature rejection logic based on hash

lubux:
The hash rejection is handled by the lower level library `go-crypto`.

### DIFF
--- a/crypto/signature.go
+++ b/crypto/signature.go
@@ -2,7 +2,6 @@ package crypto
 
 import (
 	"bytes"
-	"crypto"
 	"fmt"
 	"time"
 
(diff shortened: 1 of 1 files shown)


crypto/signature.go
```diff
@@ -305,7 +297,7 @@ func createVerifyResult(
 			signatureError = newSignatureNoVerifier()
 		case signature.SignatureError != nil:
 			signatureError = newSignatureFailed(signature.SignatureError)
-		case signature.CorrespondingSig == nil || !isHashAllowed(signature.CorrespondingSig.Hash):
+		case signature.CorrespondingSig == nil:
 			signatureError = newSignatureInsecure()
```
twiss, line 301: I think this entire case and the `newSignatureInsecure()` error can be removed entirely

lubux: The context check afterward could panic with a null pointer exceptions if we are not checking for nil. However, we could move it to signature the context check logic to make it clearer?

twiss: Yeah, I'd do that :+1:

lubux: Thanks for the input, fixed with:
https://github.com/ProtonMail/gopenpgp/pull/303/commits/274f52599ea3a90f8b3d2742d266cf48f780250c

Issue #282, opened 2024-06-06
Expose config option to decrypt unauthenticated messages


julianosk:
As other PGP libraries, the option to allow the decryption of unauthenticated messages should be exposed from the `crypto.KeyRing.Decrypt*` methods. 

There is already an option on the `go-crypto/openpgp/packet/Config` struct: `InsecureAllowUnauthenticatedMessages`, but there is no way of setting it to `true`.

PR #314, opened 2024-11-15
V3 - Add a option to allow disabling the unauthenticated messages check

JasonQuinn:
This relates to #282

### DIFF
--- a/crypto/decryption_handle_builder.go
+++ b/crypto/decryption_handle_builder.go
@@ -154,6 +154,15 @@ func (dpb *DecryptionHandleBuilder) DisableAutomaticTextSanitize() *DecryptionHa
 	return dpb
 }
 
+// DisableUnauthenticatedMessagesCheck enables to read
+// encrypted messages without Modification Detection Code (MDC).
+// MDC is mandated by the latest standard and has long been implemented
+// in most OpenPGP implementations. Messages without MDC are considered unnecessarily
+// insecure and should be prevented whenever possible.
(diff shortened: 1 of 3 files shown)


crypto/decryption_handle_builder.go
```diff
@@ -154,6 +154,11 @@ func (dpb *DecryptionHandleBuilder) DisableAutomaticTextSanitize() *DecryptionHa
 	return dpb
 }
 
+func (dpb *DecryptionHandleBuilder) DisableUnauthenticatedMessagesCheck() *DecryptionHandleBuilder {
```
lubux, line 157:
```suggestion
// DisableUnauthenticatedMessagesCheck enables to read
// encrypted messages without Modification Detection Code (MDC).
// MDC is mandated by the latest standard and has long been implemented
// in most OpenPGP implementations. Messages without MDC are considered unnecessarily
// insecure and should be prevented whenever possible.
// In case one needs to deal with messages from very old OpenPGP implementations, there
// might be no other way than to tolerate the missing MDC. Setting this flag, allows this
// mode of operation. It should be considered a measure of last resort.
// SECURITY HAZARD: Use with care.
func (dpb *DecryptionHandleBuilder) DisableUnauthenticatedMessagesCheck() *DecryptionHandleBuilder {
```

lubux (commented):
Thanks for the PR. Since this flag compromises security, appropriate documentation should be present.

PR #315, opened 2024-11-25
Add decryption option to allow disabling the integrity tag requirement (#314)

lubux:
Allows to decrypt legacy messages that have no integrity protection. This option should be used with extreme care as it allows for downgrade attacks.


lubux:
Upstream [Merges #](https://github.com/ProtonMail/gopenpgp/pull/314) into main

twiss (approved):

PR #321, opened 2024-12-10
Allow passing a message size hint

twiss:
Also, set the AEAD chunk size to the message size hint (or the actual message size, when known).

This is useful to reduce the memory usage when en/decrypting AEAD message once https://github.com/ProtonMail/go-crypto/pull/259 is merged.


crypto/encryption.go
```diff
@@ -3,13 +3,18 @@ package crypto
 import "github.com/ProtonMail/go-crypto/openpgp/packet"
 
 type EncryptionProfile interface {
-	EncryptionConfig() *packet.Config
+	EncryptionConfig(messageSizeHint uint64) *packet.Config
```
lubux, line 6: We might need more options here in the future without breaking backward compatibility.
What about introducing a `struct EncHandleSettings`  and passing `&encHandleSettings` instead with default nil?

lubux (approved):

PR #331, opened 2025-04-09
feat: Improve API for Encryption with a session key

lubux:
With RFC 9580, clients can encrypt data using AEAD with SEIPDv2 packets. However, SEIPDv2 packets are only compatible with PKESK v6 and SKESK v6 packets when the session key is encrypted using an OpenPGP certificate.

This PR improves the API to reduce the risk of encrypting data with SEIPDv2 while using a session key encrypted in a packet version below v6. Specifically, the session key now includes an indicator of whether it is intended for AEAD use or not and affects the produced packets.

### DIFF
--- a/crypto/sessionkey.go
+++ b/crypto/sessionkey.go
@@ -19,7 +19,8 @@ type SessionKey struct {
 	// Algo defines the symmetric encryption algorithm used with this key.
 	// Only present if the key was not parsed from a v6 packet.
 	Algo string
-	// v6 is a flag to indicate that the session key was parsed from a v6 PKESK or SKESK packet
+	// v6 is a flag to indicate that the session key is capable
+	// to be used in v6 PKESK or SKESK, and SEIPDv2 packets
 	v6 bool
 }
 
@@ -66,8 +67,8 @@ func (cr checkReader) Read(buf []byte) (int, error) {
 // with this SessionKey.
 // Not supported in go-mobile clients use sk.GetCipherFuncInt instead.
 func (sk *SessionKey) GetCipherFunc() (packet.CipherFunction, error) {
-	if sk.v6 {
-		return 0, errors.New("gopenpgp: no cipher function available for a v6 session key")
+	if !sk.hasAlgorithm() {
+		return 0, errors.New("gopenpgp: no cipher function available for the session key")
 	}
 	cf, ok := symKeyAlgos[sk.Algo]
 	if !ok {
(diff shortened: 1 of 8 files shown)


crypto/sessionkey.go
```diff
@@ -136,6 +190,16 @@ func NewSessionKeyFromToken(token []byte, algo string) *SessionKey {
 	}
 }
 
+// ImportAeadSessionKeyFromToken creates a SessionKey struct with the given token and algorithm,
+// which can be used in v6 PKESK or SKESK, and SEIPDv2 packets.
+func ImportAeadSessionKeyFromToken(token []byte, algo string) *SessionKey {
+	return &SessionKey{
+		Key:  clone(token),
+		Algo: algo,
+		v6:   true,
```
lubux, line 199: Instead of a boolean, we should directly encode the aead mode.

twiss: IMHO, it's not necessary as nothing will break if you use the session key with different AEAD modes, thanks to the HKDF step in SEIPDv2

crypto/sessionkey.go
```diff
@@ -136,6 +190,16 @@ func NewSessionKeyFromToken(token []byte, algo string) *SessionKey {
 	}
 }
 
+// ImportAeadSessionKeyFromToken creates a SessionKey struct with the given token and algorithm,
+// which can be used in v6 PKESK or SKESK, and SEIPDv2 packets.
+func ImportAeadSessionKeyFromToken(token []byte, algo string) *SessionKey {
```
twiss, line 195: I don't have very strong feelings about this but to be consistent with the other function, perhaps we should name it `NewAeadSessionKeyFromToken`?

And, should we have a `useAead` parameter, like `SerializeSymmetricKeyEncryptedAEADReuseKey` has? (Even though `NewAeadSessionKeyFromToken(.., false)` sounds a bit contradictory.. But it might make it easier to use this function if you want to e.g. pass `somePacket.Version == 6` or so.)

twiss (approved):

crypto/sessionkey.go
lubux: I renamed it `ImportAeadSessionKeyFromToken` because go-mobile uses all New functions as constructors and this causes it to fail because of similar arguments. However, with a boolean argument it would be possible and cleaner 👍 I gonna change that.

lubux: Done.

PR #343, opened 2025-05-23
Prepare release v2.9.0

lubux:


crypto/key_clear.go
```diff
@@ -100,12 +100,6 @@ func clearRSAPrivateKey(rsaPriv *rsa.PrivateKey) error {
 	clearBigInt(rsaPriv.Precomputed.Dp)
 	clearBigInt(rsaPriv.Precomputed.Dq)
 
-	for idx := range rsaPriv.Precomputed.CRTValues {
-		clearBigInt(rsaPriv.Precomputed.CRTValues[idx].Exp)
-		clearBigInt(rsaPriv.Precomputed.CRTValues[idx].Coeff)
-		clearBigInt(rsaPriv.Precomputed.CRTValues[idx].R)
-	}
```
twiss, line 107: Can we replace this with an assertion that `len(rsaPriv.Precomputed.CRTValues) == 0`? I think it should always be the case but it would be nice to make sure.

crypto/base_test.go
```diff
@@ -55,12 +55,6 @@ func assertRSACleared(t *testing.T, rsaPriv *rsa.PrivateKey) {
 	assertBigIntCleared(t, rsaPriv.Precomputed.Qinv)
 	assertBigIntCleared(t, rsaPriv.Precomputed.Dp)
 	assertBigIntCleared(t, rsaPriv.Precomputed.Dq)
-
-	for idx := range rsaPriv.Precomputed.CRTValues {
-		assertBigIntCleared(t, rsaPriv.Precomputed.CRTValues[idx].Exp)
-		assertBigIntCleared(t, rsaPriv.Precomputed.CRTValues[idx].Coeff)
-		assertBigIntCleared(t, rsaPriv.Precomputed.CRTValues[idx].R)
-	}
```
twiss, line 63: Same here

crypto/key_clear.go
lubux: We can also just leave it and add a lint exception 👍

twiss (approved):

